PR #15, opened 2021-11-08
feat(custom-distribution-service) define latest as the default docker image tag

dduportal:


charts/custom-distribution-service/values.yaml
```diff
@@ -5,8 +5,8 @@
 replicaCount: 1
 image:
   repository: jenkinsciinfra/custom-distribution-service
-  tag: "0.0.5"
-  pullPolicy: IfNotPresent
+  tag: "latest"
+  # pullPolicy: IfNotPresent
```
lemeurherve, line 9: Why keeping this line?

lemeurherve (approved):

charts/custom-distribution-service/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 appVersion: 0.0.5
```
lemeurherve, line 2:
```suggestion
```

lemeurherve (approved):

PR #20, opened 2021-11-09
feat(chatbot) use tag 'latest' as default

dduportal:
Signed-off-by: Damien Duportal <damien.duportal@gmail.com>

### DIFF
--- a/charts/chatbot/values.yaml
+++ b/charts/chatbot/values.yaml
@@ -6,8 +6,6 @@ replicaCount: 1
 
 image:
   repository: jenkinsciinfra/ircbot
-  tag: 165-buildbc1736
-  pullPolicy: IfNotPresent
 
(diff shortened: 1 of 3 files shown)


charts/chatbot/values.yaml
```diff
@@ -6,8 +6,7 @@ replicaCount: 1
 
 image:
   repository: jenkinsciinfra/ircbot
-  tag: 165-buildbc1736
-  pullPolicy: IfNotPresent
+  tag: latest
```
lemeurherve, line 9:
```suggestion
```
To be coherent with https://github.com/jenkins-infra/public-charts/blob/main/charts/accountapp/values.yaml#L7-L8

lemeurherve (changes requested):

lemeurherve (approved):

PR #32, opened 2021-12-09
fix: align github parameter names with the ones in plugin-site

lemeurherve:


halkeye:
yea okay, i named everything else with _, but [secrets ](https://github.com/jenkins-infra/charts-secrets/blob/main/config/plugin-site-issues/secrets.yaml#L3 )is camel case, so 🤦

halkeye:
do I need to bump chart version number?

halkeye:
actually lets fix secrets

lemeurherve:
> actually lets fix secrets

Good for me :)

charts/plugin-site-issues/templates/secret.yaml
```diff
@@ -10,5 +10,5 @@ data:
   jira_url: {{ required "jira.url is required" .Values.jira.url | toString | b64enc }}
   jira_username: {{ required "jira.username is required" .Values.jira.username | toString | b64enc }}
   jira_password: {{ required "jira.password" .Values.jira.password | toString | b64enc }}
-  github_app_id: {{ required "github.app_id is required" .Values.github.app_id | toString | b64enc }}
-  github_app_private_key: {{ required "github.app_id is required" .Values.github.app_private_key | toString | b64enc }}
+  github_app_id: {{ required "github.appId is required" .Values.github.appId | toString | b64enc }}
+  github_app_private_key: {{ required "github.appPrivateKey is required" .Values.github.appPrivateKey | toString | b64enc }}
```
lemeurherve, line 14: @halkeye the field used in the required message need to be replaced

halkeye: Yea I stared to fix secrets when I saw that. 

What's the convention? If no convention let's just update this chart and everything in else should work

lemeurherve: No convention afaik, and since the chart have to be updated I would fix all there instead of in charts-secrets. (And keep the same values fields names alongs related charts)

PR #37, opened 2021-12-17
chore: retrieve the review guidelines from its previous repository 'charts' (before its splitting and renaming in 'kubernetes-management')

lemeurherve:


dduportal (approved):
🚢

PR #87, opened 2022-01-18
chore(updatecli) ensure that only nginx stable version is tracked

dduportal:
Signed-off-by: Damien Duportal <damien.duportal@gmail.com>

### DIFF
--- a/updatecli/updatecli.d/nginx.yaml
+++ b/updatecli/updatecli.d/nginx.yaml
@@ -1,0 +1,98 @@
+title: Bump nginx:stable docker image version
+
+scms:
+  default:
+    kind: github
+    spec:
+      user: "{{ .github.user }}"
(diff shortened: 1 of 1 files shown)


lemeurherve:
Shouldn't we update nginx docker image version only here and not in jenkins-infra/kubernetes-management#1936?

dduportal:
> Shouldn't we update nginx docker image version only here and not in jenkins-infra/kubernetes-management#1936?
Perhaps it's the goal of this PR?

Good point. WDYT if we track on both, and once all are up to date, we remove the "config overrid" in kubernetes-management (e.g. removing the tag value and keep chart's defaults) ?

lemeurherve:
LGTM :)

lemeurherve (approved):

PR #94, opened 2022-02-14
feat(jenkins-kubernetes-agents) allow inserting a dockerhub secret credential for imagepullpolicy

dduportal:


charts/jenkins-kubernetes-agents/values.yaml
```diff
@@ -0,0 +1,7 @@
+## Uncomment (and fill) to add custom imaePullSecrets
```
lemeurherve, line 1: I would have uncommented all and set enabled to false.
Will do another pr :)

PR #96, opened 2022-02-15
chore typo

smerle33:


dduportal:
Many thanks, good catch!

However it's superseeded by #95 (where the whole comment was deleted).

PR #107, opened 2022-04-01
feat(rating): new release with boolean flag for readynessProbe and single quotes attributes

smerle33:
add a boolean flag to disable readynessProbe from the helmchart
and switch double quotes to single ones for php strings : https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.single

### DIFF
--- a/charts/rating/values.yaml
+++ b/charts/rating/values.yaml
@@ -70,6 +70,10 @@ tolerations: []
 
 affinity: {}
 
+#flag to disable readinessProbe in deployment.yaml L38
+readinessProbe:
+  enabled: false
+
 database:
   username: rating
   password: s3cr3t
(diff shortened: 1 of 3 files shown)


charts/rating/values.yaml
```diff
@@ -70,6 +70,10 @@ tolerations: []
 
 affinity: {}
 
+#flag to disable readinessProbe in deployment.yaml L38
+readinessProbe:
+  enabled: false
```
dduportal, line 75:
```suggestion
  enabled: true
```
This should be enabled by default. Disabling probes should be a voluntary act :)

dduportal:
The Jenkins pipeline fails with an error `ERROR: Could not find credentials entry with ID 'updatecli-github-token'
` which is not due to your change, but it is due to me (because of https://github.com/jenkins-infra/helpdesk/issues/2840).

lemeurherve (approved):

dduportal (approved):
LGTM 👍

PR #112, opened 2022-04-04
fix(jenkins-jobs) correct SSH credentials + folder jobs syntax

### LINKED ISSUE #104: (jenkins-jobs) SSH keys credentials are not working
When an SSH private key is specified as a job-level credential, it triggers a job-dsl error with the carrets (`-`)  of the header.

Should be easy to fix by using the same base64 trick as the file credentials

dduportal:
This PR closes #104, and also fix a bug introduced in #111 around folder jobs.

### DIFF
--- a/charts/jenkins-jobs/templates/_helpers.tpl
+++ b/charts/jenkins-jobs/templates/_helpers.tpl
@@ -342,7 +342,8 @@ configNode << 'com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPri
   {{- end }}
   usernameSecret({{ .usernameSecret | default false}})
   privateKeySource(class:"com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey\$DirectEntryPrivateKeySource") {
-    privateKey('{{ .privateKey }}')
+    privateKey('''{{ .privateKey }}''')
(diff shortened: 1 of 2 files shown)


charts/jenkins-jobs/templates/_helpers.tpl
```diff
@@ -342,7 +342,8 @@ configNode << 'com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPri
   {{- end }}
   usernameSecret({{ .usernameSecret | default false}})
   privateKeySource(class:"com.cloudbees.jenkins.plugins.sshcredentials.impl.BasicSSHUserPrivateKey\$DirectEntryPrivateKeySource") {
-    privateKey('{{ .privateKey }}')
+    privateKey('''{{ .privateKey }}''')
```
lemeurherve, line 345: Can't `quote`  or `squote` be usefull here?

lemeurherve (approved):

charts/jenkins-jobs/templates/_helpers.tpl
dduportal: If it handles triple quote yes 😅(i don’t like quotes though. Not clear enough)

timja (approved):

PR #120, opened 2022-04-14
feat(jenkins-jobs) change github checks defaults + allow customization

dduportal:
This PR follows up https://github.com/jenkins-infra/helpdesk/issues/2884 and introduce the following changes:

- GitHub Checks are disabled by default (only pipeline status is reported)
  - You can enable them by specifying the new attribute `enableGitHubChecks: true`
  - The former attribute `disableGitHubChecks` is deprecated (and ignored)

- GitHub Checks default name is back to `jenkins`
  - You can customize it by specifying the new name with the new `githubCheckName` attribute

### DIFF
--- a/charts/jenkins-jobs/templates/_helpers.tpl
+++ b/charts/jenkins-jobs/templates/_helpers.tpl
@@ -130,12 +130,13 @@ multibranchPipelineJob('{{ .fullId | default .id }}') {
           traits {
             gitHubSCMSourceStatusChecksTrait {
               // Note: changing this name might have impact on github branch protections if they specify status names
-              name('[infra.ci.jenkins.io] {{ .name }}')
+              name({{ .githubCheckName | default "jenkins" | squote }})
               skip({{ .disableGitHubChecks | default "false" }})
               // If this option is checked, the notifications sent by the GitHub Branch Source Plugin will be disabled.
               skipNotifications(false)
               skipProgressUpdates(false)
               // Default value: false. Warning: risk of secret leak in console if the build fails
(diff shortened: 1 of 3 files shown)


charts/jenkins-jobs/templates/_helpers.tpl
```diff
@@ -130,12 +130,17 @@ multibranchPipelineJob('{{ .fullId | default .id }}') {
           traits {
             gitHubSCMSourceStatusChecksTrait {
               // Note: changing this name might have impact on github branch protections if they specify status names
-              name('[infra.ci.jenkins.io] {{ .name }}')
-              skip({{ .disableGitHubChecks | default "false" }})
+              name({{ .githubCheckName | default "jenkins" | squote }})
+              {{- if empty .enableGitHubChecks }}
```
lemeurherve, line 134: Testing `if .enableGitHubChecks` isn't enought here I suppose? Would have been simplier than `skip({{ not .enableGitHubChecks }})`

lemeurherve (approved):
LGTM otherwise

charts/jenkins-jobs/templates/_helpers.tpl
dduportal: - In term of expctation, if you want to enable the github checks, then you want `skip(false)` in the generated job-dsl config => that is the reason of the `not`
- And `{{- if .enableGitHubChecks }}` fails if the attribute is unspecified (like "parsing error").
- And of course `{{- if or (empty .enableGitHubChecks) (not .enableGitHubChecks) }}` also fails if the attribute is unspecified

PR #132, opened 2022-05-23
feat(redirector): httpredirector chart to handle any domain redirection

smerle33:
as per https://github.com/jenkins-infra/helpdesk/issues/2943 
prepare a chart to handle the redirections


charts/httpredirector/templates/_helpers.tpl
```diff
@@ -0,0 +1,62 @@
+{{/*
```
lemeurherve, line 1: I would delete this file as none of these helpers are used in the chart.

charts/httpredirector/tests/withvalues_test.yaml
```diff
@@ -0,0 +1,23 @@
+suite: default tests
+templates:
+  - ingress.yaml
+tests:
+  - it: should generate a config map with default values
```
lemeurherve, line 5:
```suggestion
  - it: should generate a config map with custom values
```

lemeurherve (approved):

dduportal (approved):
👍

PR #162, opened 2022-07-06
fix(javadoc/jenkins.io/reports): specify secretNamespace in README examples

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/3037


lemeurherve:
These are azureFile volumes, not PV/PVC, they don't need this secretNamespace value.

See https://github.com/jenkins-infra/kubernetes-management/pull/2571#issuecomment-1176306039

PR #175, opened 2022-08-18
feat: nginx-proxy chart

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/2752


charts/repo-proxy/templates/hpa.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.autoscaling.enabled }}
+apiVersion: autoscaling/v2beta1
+kind: HorizontalPodAutoscaler
```
dduportal, line 3: Is there a reason to use an HPA for this service?

charts/repo-proxy/templates/deployment.yaml
```diff
@@ -0,0 +1,61 @@
...
+        {{- include "repo-proxy.selectorLabels" . | nindent 8 }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "repo-proxy.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      containers:
```
dduportal, line 30: Next step: add a dynmaic persistent volume (not an external PV + PVC, but a dynamic associated to each pod of the deployement)

charts/repo-proxy/templates/hpa.yaml
lemeurherve: None, I've used `helm create repo-proxy` to generate the chart, this was included.

charts/repo-proxy/templates/deployment.yaml
lemeurherve: This kind? https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/
(Needs a PVC though)

lemeurherve:
Tested locally with Rancher:
image

~/.m2/settings.xml:
```
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" 
  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
  <mirrors>
    <!-- <mirror>
      <id>maven-cache-central</id>
      <url>http://HOSTNAME/mirrors/central</url>
      <mirrorOf>central</mirrorOf>
    </mirror>
    <mirror>
      <id>maven-cache-jenkins</id>
      <url>http://HOSTNAME/mirrors/jenkins</url>
      <mirrorOf>repo.jenkins-ci.org</mirrorOf>
    </mirror> -->
    <mirror>
        <id>azure</id>
        <url>http://localhost:8000/public/</url>
        <mirrorOf>repo.jenkins-ci.org</mirrorOf>
    </mirror>
    <!-- <mirror>
        <id>azure-incrementals</id>
        <url>https://repo.azure.jenkins.io/incrementals/</url>
        <mirrorOf>incrementals</mirrorOf>
    </mirror> -->
  </mirrors>
</settings>
```

charts/nginx-proxy/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
```
dduportal, line 3: Do we need to create a serviceaccount?

charts/nginx-proxy/values.yaml
```diff
@@ -0,0 +1,96 @@
...
+  #   cpu: 100m
+  #   memory: 128Mi
+
+autoscaling:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 100
+  targetCPUUtilizationPercentage: 80
+  # targetMemoryUtilizationPercentage: 80
+
```
dduportal, line 77: What is the purpose of these values?

dduportal (approved):

charts/nginx-proxy/values.yaml
lemeurherve: These are the default values generated by helm for HPA, I missed them when I've deleted the hpa.yaml

charts/nginx-proxy/templates/serviceaccount.yaml
lemeurherve: It's by default in the chart generated by helm. As long as we keep the corresponding value to falsy it won't be created.

dduportal: I remember that, unless specifying a custom svcaccount, the svcaccount `default` is used on the same namespace.

Ideally, could we specify "no svcaccount" so that any request to the Kube API from this pod would be denied unless configured with `kubeconfig` ?

lemeurherve: I've set serviceAccount.create to false by default.
Do you prefer I remove the serviceaccount.yaml (and associated values) instead?

lemeurherve: Removed

PR #208, opened 2022-09-13
fix(artifact-caching-proxy): health probes & statefulset instead of deployment

lemeurherve:


dduportal:
Can you update the PR title (or split it)? The change deployment to statefulset is worth a word ;)

PR #212, opened 2022-09-15
feat: create `plugin-health-scoring` helm chart

alecharp:
Related to https://github.com/jenkins-infra/helpdesk/issues/3114.

I created this Helm Chart using `helm create` and modified `values.yml`, `Chart.yml` and `deployment.yml`.


alecharp:
At this state, I created a local cluster using `k3d`, installed the `bitnami/postgresql` helm chart, validated the DB availability and then installed this chart (with some `--set` for the database and GitHub OAuth token). The application started correctly, I could access it (once forwarding the port). 
Note: for the moment, the docker image is built on `ci.jenkins.io`, so its name is `jenkins4eval/plugin-health-scoring`. As discussed with @dduportal and @lemeurherve, we should move the ci to the infra instance and then the image name in this chart would be the correct one.

dduportal (approved):
LGTM 👍 Let's go with this first version!

PR #236, opened 2022-10-06
fix: restore wiki 0.3.30 in the helm chart repo index

lemeurherve:
There has been a race condition after the successive merge of #235 and #234, resulting in this commit where we can see only accountapp remained in the helm charts repo index.yaml: https://github.com/jenkins-infra/helm-charts/commit/5aca4f0e811ed4c0a10851637b828b731bc33e24

### DIFF
--- a/index.yaml
+++ b/index.yaml
@@ -1612,6 +1612,16 @@ entries:
   wiki:
   - apiVersion: v2
     appVersion: 0.0.7
+    created: "2022-10-06T09:39:56.585680469Z"
+    description: A Helm chart for wiki static content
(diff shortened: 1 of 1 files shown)


lemeurherve:
> There has been a race condition after the successive merge of #235 and #234, resulting in this commit where we can see only accountapp remained in the helm charts repo index.yaml: [5aca4f0](https://github.com/jenkins-infra/helm-charts/commit/5aca4f0e811ed4c0a10851637b828b731bc33e24)

We should fix the build to allow safe parallel chart releases, then ensure each last chart release is effectively present in https://github.com/jenkins-infra/helm-charts/blob/gh-pages/index.yaml

PR #248, opened 2022-10-13
fix(artifact-caching-proxy): chown lost+found too

lemeurherve:


charts/artifact-caching-proxy/templates/statefulset.yaml
```diff
@@ -37,7 +37,7 @@ spec:
           runAsUser: 0
           runAsGroup: 0
         command: ['sh', '-c']
-        args: ["chown -Rv {{ .Values.initContainer.chownArgs }}  {{ .Values.paths.cachePath }} && ls -l {{ .Values.paths.cachePath }}"]
```
dduportal, line 40: Not sure what does it change?

Previous version uses the `-R` flag (e.g. "recursive"). So *when* the command was triggered, if `lost+found` was present, then it would have been treated.

lemeurherve: Yup, you confirmed my suspicions, closing this PR.

PR #250, opened 2022-10-14
Add updatecli for ircbot

halkeye:


halkeye:
I think its failing because updatecli is trying to apply the diff to master, not the current branch. Thinking might need to be two different PRs

hervelemeur:
Yeah, all updatecli manifest introductions are failing, nonworries no need for 2 PRs, we'll merge it as it is.
Thanks for the PR!

lemeurherve (approved):

PR #252, opened 2022-10-14
fix(artifact-caching-proxy): better initContainers command

lemeurherve:
Trying to debug more easily my issue with the /var/cache/nginx/lost+root owner as root.

### DIFF
--- a/charts/artifact-caching-proxy/values.yaml
+++ b/charts/artifact-caching-proxy/values.yaml
@@ -70,7 +70,7 @@ persistence:
   accessMode: ReadWriteOnce
 
 initContainers:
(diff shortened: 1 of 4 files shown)


charts/artifact-caching-proxy/values.yaml
```diff
@@ -72,10 +68,13 @@ persistence:
   size: 50
   storageClass: ""
   accessMode: ReadWriteOnce
+
+initContainers:
+  enabled: false
```
lemeurherve, line 73:
```suggestion
  enabled: true
```

dduportal (approved):

PR #254, opened 2022-10-15
Bump ircbot image to 1.9.0

NotMyFault:
Restores logging functionality.

### DIFF
--- a/charts/ircbot/Chart.yaml
+++ b/charts/ircbot/Chart.yaml
@@ -4,4 +4,4 @@ name: ircbot
 maintainers:
   - name: timja
   - name: slide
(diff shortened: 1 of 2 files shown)


hervelemeur:
FYI there is an updatecli manifest taking care of these updates automatically: https://github.com/jenkins-infra/helm-charts/pull/255

NotMyFault:
> FYI there is an updatecli manifest taking care of these updates automatically: #255

Yeah, I was aware, but I wasn't too sure _when_ it does its magic, hence I created this PR yesterday.

But let's rather use the updatecli one, bonus points for formatting :P

PR #279, opened 2022-10-25
feat(artifact-caching-proxy): expose healthcheck path without auth

lemeurherve:
Ref: https://github.com/jenkins-infra/helpdesk/issues/2752


dduportal (approved):
👍

PR #282, opened 2022-10-26
fix(rss2twitter): quotes around boolean env vars and runAsUser

lemeurherve:


charts/rss2twitter/values.yaml
```diff
@@ -24,7 +24,7 @@ securityContext:
     drop:
     - ALL
   readOnlyRootFilesystem: true
-  runAsNonRoot: true
```
lemeurherve, line 27: runAsNonRoot only works with a numerical ID.

dduportal: Nope: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#podsecuritycontext-v1-core.

It is a boolean. You are confusing this option with `runAsUser`

lemeurherve: What I mean is runAsNonRoot is testing for a numerical id in Kubernetes code, and as the rss2twitter defines its user with `USER app` the deployment fails with `runAsNonRoot`.

https://stackoverflow.com/a/49729786/4074148

PR #297, opened 2022-11-04
feat(artifact-caching-proxy): add response time to nginx logs

lemeurherve:
From https://jenkins.datadoghq.com/integrations/nginx

> Note: The default NGINX log format does not have a request response time.


dduportal (approved):

dduportal:
Looks good as per nginx doc in https://nginx.org/en/docs/http/ngx_http_log_module.html

> $request_time
request processing time in seconds with a milliseconds resolution; time elapsed between the first bytes were read from the client and the log write after the last bytes were sent to the client

PR #300, opened 2022-11-04
fix(artifact-caching-proxy): always add annotations other than podAnnotations

lemeurherve:
The additional ones were never declared.


smerle33 (dismissed):
🚀

smerle33 (approved):
🚀

PR #303, opened 2022-11-07
feat(artifact-caching-proxy) use feature flags to configure datadog integration

dduportal:
This PR introduces the following changes:

- 🐛 correct the datadog annotation for metrics, following https://docs.datadoghq.com/containers/kubernetes/integrations/?tab=kubernetesadv2#configuration (comments added)

- 🧹 2 values removed:
  - `logCollection.enabled` => replaced by `datadog.logCollection.enabled` (see below)
  - `logCollection.config` => moved into the template as one of the values requires to be set to the container name (which is currently `{{ .Charts.Name }}`

- ✅ 2 new values added to allow/disallow integration with datadog (enabled by default):
  - `datadog.logCollection.enabled`
  - `datadog.metricsCollection.enabled`


dduportal:
Please note that it works as expected: tested manually on `prodpublick8s` and the Datadog's Nginx integration started to get data:

Capture d’écran 2022-11-07 à 13 55 13

Capture d’écran 2022-11-07 à 14 12 51

charts/artifact-caching-proxy/templates/statefulset.yaml
```diff
@@ -27,9 +32,13 @@ spec:
...
-        {{- if .Values.logCollection.enabled }}
-        # log collection with Datadog
-        ad.datadoghq.com/{{ .Chart.Name }}.logs: {{ .Values.logCollection.config | squote }}
+        {{- end }}
+        {{- if .Values.datadog.logCollection.enabled }}
+        # Nginx Log collection with Datadog:
+        # - https://docs.datadoghq.com/integrations/nginx/?tab=kubernetes#log-collection
+        # - https://docs.datadoghq.com/containers/kubernetes/log/?tab=kubernetes#configuration
+        #   1. With 'ad.datadoghq.com/xxx.logs' , the 'xxx' must be the name of the pod's container name
+        ad.datadoghq.com/{{ .Chart.Name }}.logs: '[{"source":"nginx","service":"{{ .Chart.Name }}","tag": "env:prod"}]'
```
lemeurherve, line 41: This change bothers me a bit as I think it should be configurable depending on where you deploy it. Ex: the tags.

charts/artifact-caching-proxy/values.yaml
```diff
@@ -58,10 +58,12 @@ resources: {}
...
 #   memory: 128Mi
 
-# log collection with Datadog
-logCollection:
-  enabled: true
-  config: '[{"source":"nginx","service":"artifact-caching-proxy","tag": "env:prod"}]'
+# Datadog Integration
+datadog:
+  metricsCollection:
+    enabled: true
```
lemeurherve, line 64: Is there a case where we don't want to enable metrics collection?
Initially I've put a logCollection.enabled as by default it's disabled (cf datadog doc), but it's not the case for metrics.

charts/artifact-caching-proxy/templates/statefulset.yaml
dduportal: I propose to open a subsequent PR to allow specifying the tags as a value.

But for now, the "raw" specification of the 0.4.4 version is a bug from my view: if the template changes the container names, then the chart installation are broken.

lemeurherve, line 41:
```suggestion
        ad.datadoghq.com/{{ .Chart.Name }}.logs: '[{"source":"nginx","service":"{{ .Chart.Name }}"}]'
```

charts/artifact-caching-proxy/values.yaml
dduportal: I don't see a direct case indeed, as much as I don't see why log collection should be disabled here: datadog documentation states that log collection is disabled by default_on the datadog agent_. Switching the flag on/off in this chart does not impact the datadog agent configuration.

That's why I chose to have the same feature flag for both: we might want to disable it in the future.

(I though about cleaning up annotations if needed, due to an admission controller. But for statefulset/deployment I don't see the case, compared to ingress annotation that could be refused by the ingress admission)

charts/artifact-caching-proxy/templates/statefulset.yaml
dduportal: Looks good: tag specification should be configurable through value as you mentioned.

PR #356, opened 2022-12-16
fix(plugin-health-scoring): checksum on secret content and test if not empty only

lemeurherve:


charts/plugin-health-scoring/templates/deployment.yaml
```diff
@@ -16,7 +16,7 @@ spec:
         app.kubernetes.io/name: {{ include "plugin-health-scoring.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
       annotations:
-        checksum/config: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+        checksum/config: {{ print .Values.database.password .Values.config.githubOauth | sha256sum }}
```
dduportal, line 19: oh I see, it's not easy to access the `data` directive in the `secret.yaml` rendered template.
🤔 I fear that this line would be considered "fragile" as it requires a manual update each time a new attribute is added. Not sure what would be the best (or the least worst)

lemeurherve: Should be addressed in [jenkins-infra/helm-charts@`aa6813d`](https://github.com/jenkins-infra/helm-charts/pull/356/commits/aa6813d617dc142207114d76daa9c1eb72a26783) 🙂

lemeurherve:
No intermittent error anymore 🙂

PR #402, opened 2023-01-25
fix(GHA): add GITHUB_TOKEN

lemeurherve:
I've noticed this warning on actions, which are (too) frequently failing to download helm-unittest:

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,7 +28,7 @@ jobs:
         uses: azure/setup-helm@v3
 
       - name: install helm unittests
-        with:
+        env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           helm env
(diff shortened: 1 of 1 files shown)


lemeurherve:
As the corresponding `JENKINS_ADMIN_APP_PRIVKEY` secret exists at the org level I don't understand the following error:

> Error: Input required and not supplied: private_key

dduportal (changes requested):
- Secrets are not available in Pull Requests by default (unless you set up the correct set of events and settings for trusted authors - https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/) so you won't be able to do anything.
- However, the variable `GITHUB_TOKEN` is always set in a PR: you should be able to use it by following what the error says: specify it in the `env:` section of the action helm

dduportal (approved):

.github/workflows/test.yml
```diff
@@ -28,6 +28,8 @@ jobs:
         uses: azure/setup-helm@v3
 
       - name: install helm unittests
+        with:
```
dduportal, line 31:
```suggestion
        env:
```

dduportal (approved):

PR #404, opened 2023-01-25
chore(gha) Fix helm plugin network errors and helm setup action warning

dduportal:
Fixup of #402 

- Warning in #402 's body comes from the `azure/setup-helm` (ref. https://github.com/Azure/setup-helm#example). By specifying a given version (extracted from our Docker Image principal branch with a dummy `curl | grep`), no need for token anymore
- Trying to solve the "helm plugin install" network error by using @lemeurherve 's trick of ... retrying a 2nd time "to be sure".

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,9 +36,10 @@ jobs:
       - name: install helm unittests
         run: |
           helm env
-          helm plugin install https://github.com/quintush/helm-unittest
-          # Repeat to fight against network errors in  GHA runners
-          helm plugin install https://github.com/quintush/helm-unittest
+          # Repeat 3 times to fight against network errors in  GHA runners
+          helm plugin install https://github.com/quintush/helm-unittest \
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -23,16 +23,23 @@ jobs:
...
+        with:
+          version: "${{ steps.tools-versions.outputs.CURRENT_HELM_VERSION }}"
 
       - name: install helm unittests
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           helm env
-          helm plugin install https://github.com/quintush/helm-unittest
+          # Repeat 3 times to fight against network errors in  GHA runners
```
lemeurherve, line 39: 🎉  :trollface:

lemeurherve (approved):

.github/workflows/test.yml
dduportal: We are here to do nasty hotfixes and to chew bubble gum. And we outta bubble gum

PR #411, opened 2023-02-02
fix(artifact-caching-proxy): correct nginx log format

smerle33:
to avoid a double double quotes

### DIFF
--- a/charts/artifact-caching-proxy/Chart.yaml
+++ b/charts/artifact-caching-proxy/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 description: artifact-caching-proxy is a Nginx caching proxy in front of repo.jenkins-ci.org
 name: artifact-caching-proxy
 type: application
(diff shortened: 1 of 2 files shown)


dduportal:
While reviewing, first proposal: the PR title is incorrect:

- This change, while not being a major change, is scoped to the application. Given the current title saying `correct log format`, WDYT about using `fix` instead of `chore` ?
- The `nginx/log` between parenthesis does not map to any of the "elements" on this repositories. If you look at the other PR, you'll see the pattern (e.g. we specify the chart name)

dduportal (changes requested):
LGTM for the change itself.

Blocking the merge for the 2 following elements:

- Update the PR title (based on my previous comment)
- Update the helm chart version (in `charts/artifact-caching-proxy/Chart.yaml`) otherwise the release will fail

dduportal (approved):
LGTM 👍

PR #580, opened 2023-08-07
fix(mirrorbits): use the PVC name defined in values.yaml everywhere

lemeurherve:
As it is currently, even with a `repository.name` value specified in values.yaml, the PVC is named only in the PVC template, it has a generic release name + "-binary" in the deployments templates.

This PR fixes that.

Note: I don't see why there are `$.` in front of some of the `if` in these deployments an in the PVC template 🤔

### DIFF
--- a/charts/mirrorbits/Chart.yaml
+++ b/charts/mirrorbits/Chart.yaml
@@ -5,4 +5,4 @@ maintainers:
 - email: me@olblak.com
   name: olblak
 name: mirrorbits
-version: 0.52.2
+version: 0.53.0
--- a/charts/mirrorbits/templates/deployment.files.yaml
+++ b/charts/mirrorbits/templates/deployment.files.yaml
@@ -60,6 +60,8 @@ spec:
         - name: conf
           configMap:
             name: {{ include "mirrorbits.fullname" . }}-files
+        {{ if .Values.repository.persistentVolumeClaim.enabled -}}
         - name: binary
           persistentVolumeClaim:
-            claimName: {{ include "mirrorbits.fullname" . }}-binary
+            claimName: {{ .Values.repository.name | default (printf "%s-binary" (include "mirrorbits.fullname" .)) }}
+        {{- end }}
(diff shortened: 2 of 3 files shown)


dduportal (approved):
Good catch 👍 

As this chart does not have any unit tests, I've tried an `helmfile diff` against the production cluster `publick8s` and the result looks good:

```diff
helmfile diff -f ./clusters/publick8s.yaml -l name=mirrorbits --suppress-secrets --skip-deps --context=2
Decrypting secret /Users/dadou/workspace/jenkins-infra/kubernetes-management/secrets/config/mirrorbits/secrets.yaml
Comparing release=mirrorbits, chart=../../helm-charts/charts/mirrorbits
mirrorbits, mirrorbits, Deployment (apps) has changed:
...
    labels:
      app.kubernetes.io/name: mirrorbits
-     helm.sh/chart: mirrorbits-0.52.2
+     helm.sh/chart: mirrorbits-0.53.0
      app.kubernetes.io/instance: mirrorbits
      app.kubernetes.io/version: "v0.5.2"
...
mirrorbits, mirrorbits, Ingress (networking.k8s.io) has changed:
...
    labels:
      app.kubernetes.io/name: mirrorbits
-     helm.sh/chart: mirrorbits-0.52.2
+     helm.sh/chart: mirrorbits-0.53.0
      app.kubernetes.io/instance: mirrorbits
      app.kubernetes.io/version: "v0.5.2"
...
mirrorbits, mirrorbits, Service (v1) has changed:
...
    labels:
      app.kubernetes.io/name: mirrorbits
-     helm.sh/chart: mirrorbits-0.52.2
+     helm.sh/chart: mirrorbits-0.53.0
      app.kubernetes.io/instance: mirrorbits
      app.kubernetes.io/version: "v0.5.2"
...
mirrorbits, mirrorbits-files, Deployment (apps) has changed:
...
    labels:
      app.kubernetes.io/name: mirrorbits
-     helm.sh/chart: mirrorbits-0.52.2
+     helm.sh/chart: mirrorbits-0.53.0
      app.kubernetes.io/instance: mirrorbits
      app.kubernetes.io/version: "v0.5.2"
...
mirrorbits, mirrorbits-files, Service (v1) has changed:
...
    labels:
      app.kubernetes.io/name: mirrorbits
-     helm.sh/chart: mirrorbits-0.52.2
+     helm.sh/chart: mirrorbits-0.53.0
      app.kubernetes.io/instance: mirrorbits
      app.kubernetes.io/version: "v0.5.2"
...
```

PR #585, opened 2023-08-09
fix(mirrorbits): correct service selector label

### LINKED ISSUE #589: The `files` service selector `app.kubernetes.io/name` label of mirrorbits isn't the same as the deployment one
When working on https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-1660258021 trying to deploy https://github.com/jenkins-infra/kubernetes-management/pull/4211 I've encountered 503 issue with the `files` service.

This error is due to the fact that its selector label isn't the same as the deployment label (like in the other service), it's using"mirrorbits.fullname":
https://github.com/jenkins-infra/helm-charts/blob/adc6fa5825c662e175da60df245ceca95fad21f1/charts/mirrorbits/templates/service.files.yaml#L15

While the corresponding deployment is using "mirrorbits.name":
https://github.com/jenkins-infra/helm-charts/blob/adc6fa5825c662e175da60df245ceca95fad21f1/charts/mirrorbits/templates/deployment.files.yaml#L16

Note: `get.jenkins.io` (the service for which this chart has been initialy designed) isn't affected as it does not rely on any `fullNameOverride` or `nameOverride` used by the template helper to generate this label.

lemeurherve:
Fixes #589

### DIFF
--- a/charts/mirrorbits/Chart.yaml
+++ b/charts/mirrorbits/Chart.yaml
@@ -5,4 +5,4 @@ maintainers:
 - email: me@olblak.com
   name: olblak
 name: mirrorbits
-version: 0.53.3
+version: 0.54.0
--- a/charts/mirrorbits/templates/service.files.yaml
+++ b/charts/mirrorbits/templates/service.files.yaml
@@ -12,5 +12,5 @@ spec:
       protocol: TCP
       name: http
   selector:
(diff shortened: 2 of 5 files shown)


dduportal (approved):
LGTM

- Fix is legit
- It adds unit tests to cover it

image

PR #593, opened 2023-08-10
fix(accountapp): add missing default `image.pullPolicy` and unit tests

lemeurherve:
This PR fixes the chart by adding the missing default `image.pullPolicy` used in its template, and add corresponding unit tests.

Ref: https://github.com/jenkins-infra/kubernetes-management/pull/4243#pullrequestreview-1572163696

> [...] the chart being buggy: https://github.com/jenkins-infra/helm-charts/blob/2aff68e63375846a9942f8ff556d70b4a6e60f9c/charts/accountapp/templates/deployment.yaml#L29 
> 
> (The helmfile diff shows that the key is still specified but its value is undefined, which does not behave as described in the kube doc).


dduportal (approved):
LGTM

(note there are two patterns that could be used here: either always specifying the attribute `imagePullPolicy`in the deployment like it's done, which require setting up the proper value system, or only specifying the whole hey + value only if the value is defined and delegating to Kubernetes default behavior. Both are fine)

lemeurherve:
https://github.com/jenkins-infra/helm-charts/pull/595#issuecomment-1674571466 OK, I'll bump the chart version and merge this PR.

PR #595, opened 2023-08-11
fixup: check-chart-versions-bump.yml

lemeurherve:
There was some errors in the initial version 🙈 

This one also allow checking all modified charts instead of stopping at the first unbumped version.


lemeurherve:
Confirmed working as intended now:
image

PR #596, opened 2023-08-11
feat: test only modified charts with unit tests on pull requests

lemeurherve:
This PR allows to test only modified charts containing unit tests on pull requests, and remove the need to specify which chart should be tested in the GitHub Action workflow file.

On the`main` branch the GitHub Action tests every chart containing unit tests.

### Testing done
Tested with 3 modified charts, 2 of them containing unit tests: (closes #597)

image

When there is no modified chart with unit tests, helm related steps are skipped:
image

Tested on `main` branch on a fork.

### Potential improvement
On the `main` branch run only on modified charts since last commit of previous push instead of running on every chart containing unit tests.

### Note
Previous status checks required in the `main` branch protection of this repo will have to be replaced by the new ones.

On this PR:
image


dduportal (commented):
LGTM

2 things (not blocking and no problem if don't find alignement as it's mostly nitpicking):


- I'm not sure to understand the value of using 2 different jobs (which are consuming more GHA minutes as it starts a worker for each job)?

> On the main branch run only on modified charts since last commit of previous push instead of running on every chart containing unit tests.

I'm not in favor of conditional tests on the main branch, under the assumptions of these are unit tests so it's ligthweight to test AND could be hit by a "something changed in a transitive dependency". 
(I wouldn't have written that for acceptance test which are usually slow and flaky so could be valuable to  conditionnaly run).

dduportal (dismissed):
sorry, misclicked: my previous review should have been an approval.

hervelemeur:
> I'm not sure to understand the value of using 2 different jobs (which are consuming more GHA minutes as it starts a worker for each job)?

I find it more elegant, more simpler separate scripts.
About the GHA minutes, we're not using a lot of GHAs accross the jenkins-infra orgs, it shouldn't be a problem I think.
I'm not attached to it, that can be changed if you think it needs to.

> > On the main branch run only on modified charts since last commit of previous push instead of running on every chart containing unit tests.
> 
> I'm not in favor of conditional tests on the main branch, under the assumptions of these are unit tests so it's ligthweight to test AND could be hit by a "something changed in a transitive dependency". (I wouldn't have written that for acceptance test which are usually slow and flaky so could be valuable to conditionnaly run).

Agreed, even if I don't see any transitive dependency in this repo affecting multiple charts. Let's keep it that way, running all tests on main.

dduportal:
New unforeseen problem: what checks should be defined as "required" in order to allow merging?

lemeurherve:
> New unforeseen problem: what checks should be defined as "required" in order to allow merging?

Good catch, didn't find any way to set required checks on "unknown" ones.

I've modified the GHA to include only one job with a for loop.

lemeurherve:
> New unforeseen problem: what checks should be defined as "required" in order to allow merging?

Note: we currently don't have all unittests checks as required as the list of tested charts grown over time.

dduportal (approved):
LGTM!

I let you merge, and then update the status check setup of the repo.

PR #598, opened 2023-08-12
fix(check-chart-versions-bump): don't fail on new chart

lemeurherve:
This PR fixes the `check-chart-versions-bump` GitHub Action workflow to not fail if the pull request adds a new chart (ie the corresponding `charts/<chart-name>/Chart.yaml` file doesn't exist in the target branch) or if the pull request Chart.yaml version isn't set.

It also refactors the workflow to use a multi jobs mechanism similar to #596 and to run only if there are changes in `./charts/**`.

Tested on a fork.

Example with a new `anewchart` chart and a change in `accountapp` chart for which the version hasn't been bumped:

image

image

> Info: charts/anewchart/Chart.yaml doesn't exist or a version can't be read from it on the target branch.
> Info: the version of the 'anewchart' chart has been bumped from '' to '1.2.3'.

### DIFF
--- a/.github/workflows/check-chart-versions-bump.yml
+++ b/.github/workflows/check-chart-versions-bump.yml
@@ -1,45 +1,66 @@
 name: Check chart version bump
 on:
   pull_request:
+    paths:
+      - 'charts/**'
 jobs:
-  check-chart-version-bump:
+  find_modified_charts:
     runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


dduportal (approved):
LGTM!

Same as https://github.com/jenkins-infra/helm-charts/pull/596#pullrequestreview-1577274083: not blocking (mostly nitpicking), but the value of using separated jobs is not clear to me.

PR #606, opened 2023-08-21
fix(check-chart-version-bump): use only one job to be able to set it as required

lemeurherve:
This PR uses only one job intead of a matrix with 2 separate jobs so this GHA check can be set as required in branch protection settings.

Ref: https://github.com/jenkins-infra/helm-charts/issues/596#issuecomment-1686243269


dduportal (approved):
Good idea: using this method allows portability to another CI system (have you heard about Jenkins? :trollface: ).


It also fixes the initial problem, nice one!

PR #609, opened 2023-08-22
feat(mirrorbits): transplant `rsyncd` service from `rsyncd-mirror` chart

lemeurherve:
As the two charts share a lot of templates and values, it makes more sense to extract the `rsyncd` capacity from the old chart to this one instead of updating the old one like this one and keep both maintained.

Ref: https://github.com/jenkins-infra/helpdesk/issues/2649


dduportal (approved):
LGTM. Diff on get.jenkins looks good, and there are unit tests 👏

dduportal:
You can proceed to merge: only the `updatecli` checks fails (as usual)

PR #627, opened 2023-08-23
chore(mirrorbits): more sensible `LoadBalancer` type for `rsyncd` service

lemeurherve:


dduportal (changes requested):
Usually, the default kind of a Kuberenetes Service in a Helm chart is `ClusterIP` because it always exists in any Kubernetes cluster.

That is not the case for `LoadBalancer`: if you don't have a LoadBalancer controller installed then no service can be of this kind and the installation would fail.


Beside, I have a personal preference of setting a kind `LoadBalancer` by overrding values because it is easier to understand the installatiobn topology (particularly in our case where we specify an annotation for the public IP)

hervelemeur:
Agreed. Can you close this PR please?

dduportal:
> Agreed. Can you close this PR and linked issue please?

yes! Safe travel buddy :)

PR #652, opened 2023-09-05
feat(gha): test each modified charts with `helm template`

lemeurherve:
As a complement to unit-tests, this PR adds a basic test running `helm template <chartName>` to catch basic errors without requiring the presence of unit-tests.

The long goal is to improve our helm chart test coverage by adding unit-tests to all of them.


dduportal (changes requested):
Not sure to understand the value of adding `helm template` over a unit test as it is literrally the goal of helm unit tests to verify this.

PR #665, opened 2023-09-06
feat!(rsyncd) harden container execution with image `1.0.0`

dduportal:
BREAKING CHANGE: former values need to be adapted

- Supports the new image [1.0.0](https://github.com/jenkins-infra/docker-rsyncd/releases/tag/1.0.0)
  - Default user `nobody`
  - New default configuration scheme
  - Changed `motd`
-  Fixed pod level attributes (tolerations, nodeselectors, security context, etc.)


charts/rsyncd/templates/deployment.yaml
```diff
@@ -37,20 +45,26 @@ spec:
...
+              readOnly: true # Default for configmaps
+              subPath: rsyncd.inc
             - name: jenkins-motd
-              mountPath: {{ .Values.configuration.motd_path }}
-              subPath: jenkins-motd
+              mountPath: {{ .Values.configuration.motd.path }}
+              subPath: jenkins.motd
+              readOnly: true
+            {{- range .Values.configuration.components }}
+            - name: rsyncd-conf
```
lemeurherve, line 62: Doesn't this name need to be unique? What happens if there is more than one component?

charts/rsyncd/templates/deployment.yaml
```diff
@@ -37,20 +45,26 @@ spec:
...
-            {{- if .Values.configurationFiles.jenkinsMotd.override }}
+              mountPath: /etc/rsyncd.d/rsyncd.inc
+              readOnly: true # Default for configmaps
+              subPath: rsyncd.inc
             - name: jenkins-motd
-              mountPath: {{ .Values.configuration.motd_path }}
-              subPath: jenkins-motd
+              mountPath: {{ .Values.configuration.motd.path }}
+              subPath: jenkins.motd
+              readOnly: true
```
lemeurherve, line 60:
```suggestion
              readOnly: true # Default for configmaps
```
Like L55 & L64

charts/rsyncd/templates/deployment.yaml
```diff
@@ -37,20 +45,26 @@ spec:
...
+            {{- range .Values.configuration.components }}
+            - name: rsyncd-conf
+              mountPath: /etc/rsyncd.d/{{ .name }}.conf
+              readOnly: true # Default for configmaps
+              subPath: {{ .name }}.conf
+            - name: datadir-{{ .name }}
+              mountPath: {{ .path }}
               readOnly: true
```
lemeurherve, line 68:
```suggestion
              readOnly: true # Default for configmaps
```
Like L55 & L64

charts/rsyncd/tests/custom_values_test.yaml
```diff
@@ -16,21 +44,190 @@ tests:
...
+          path: data["jenkins.motd"]
+          pattern: SUPER MOTD
+  - it: should create a custom rsyncd-conf config map
+    template: configmap.rsyncd-conf.yaml
+    set:
+      configuration:
+        motd:
+          path: /tmp/motd
+        components:
+          - name: jenkins
```
lemeurherve, line 172: I would add a test with 2 components.

lemeurherve (commented):
Nice! Some comments and a possible issue to fix, maybe.

charts/rsyncd/tests/custom_values_test.yaml
lemeurherve: Ref: https://github.com/jenkins-infra/helm-charts/pull/665/files#r1317715286

charts/rsyncd/templates/deployment.yaml
dduportal: The name is internal to the Deployment. it's not a resource.

charts/rsyncd/templates/deployment.yaml
dduportal: not for this one: the datadir is not a configmap

charts/rsyncd/tests/custom_values_test.yaml
dduportal: > Ref: https://github.com/jenkins-infra/helm-charts/pull/665/files#r1317715286

Unrelated (see my comment) but worth testing with 2 components yes

charts/rsyncd/templates/deployment.yaml
lemeurherve: Not a resource OK, and sorry if I'm missing something here but WDYT of adding the `name` to it so there won't be multiple `rsyncd-conf` name entries if there are multiple components (which doesn't feel right to me)?

```suggestion
            - name: rsyncd-conf-{{ name }}
```

dduportal: Your suggested change will break the deployment: the `name` attribute of a pod's container `volumeMount` is used to reference which volume of the same pod is projected to the mount.
https://kubernetes.io/docs/concepts/storage/volumes/ and https://kubernetes.io/docs/tasks/configure-pod-container/configure-volume-storage/ will help you about this topic.

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.25/#volumemount-v1-core is the reference doc about the fields, their type and goals:

> name:     This must match the Name of a Volume.


It means that if we change the name here, we also have to change it on the `volumes` section of the pod.

And the reference documentation about pod volumes states that the name must be "DNS compliant" (as usual) and must be within the pod: proof that there is no collision (as I said, it is NOT an API object but only data inside the representation of a Pod in the API).

dduportal: Additionnaly, the name `rsync-conf` is unique but it contains multiple entries in this deployment (to avoid creating  multiple configmap entries).

I wonder if we should not to the same for the motd: adding it as en entrry of this configmap so less entries per deployment

lemeurherve: Sure about the corresponding volume and volumeMount names, what's troubling me here is the fact that multiple entries of the map will have the same `name`.

I would have added the `{{ name }}`, and the same L83 + inclusion of the rsyncd-conf into the range loop.

charts/rsyncd/tests/defaults_test.yaml
```diff
@@ -1,23 +1,136 @@
...
+      # No Rsyncd component by default mounted by default
+      - notExists:
+          path: spec.template.spec.containers[0].volumeMounts[2]
+      - notExists:
+          path: spec.template.spec.volumes[2]
+      - notExists:
+          path: spec.template.spec.containers[0].volumeMounts[3]
+      - notExists:
+          path: spec.template.spec.volumes[3]
+  - it: should create a the default motd config map
```
lemeurherve, line 80:
```suggestion
  - it: should create a default motd config map
```

charts/rsyncd/tests/defaults_test.yaml
```diff
@@ -1,23 +1,136 @@
...
+          count: 1
+      - isKind:
+          of: ConfigMap
+      - equal:
+          path: metadata.name
+          value: RELEASE-NAME-rsyncd-jenkins-motd
+      - matchRegex:
+          path: data["jenkins.motd"]
+          pattern: JENKINS MIRROR
+  - it: should create a the default rsyncd-conf config map
```
lemeurherve, line 93:
```suggestion
  - it: should create a default rsyncd-conf config map
```

charts/rsyncd/templates/deployment.yaml
dduportal: > multiple entries of the map will have the same name.

This is expected: one configmap to rule them all, so same name to point to the same configmap.

Your suggestion breaks the deployments, we simply can't do it.

dduportal: More precisely: you are pointing what is the goal of the `subPath` attribute

charts/rsyncd/tests/custom_values_test.yaml
dduportal: Implemented in https://github.com/jenkins-infra/helm-charts/pull/665/commits/8ca5d5bad34d45b8793e2f67358567c01cfcbe0d

charts/rsyncd/templates/deployment.yaml
dduportal: See https://github.com/jenkins-infra/helm-charts/pull/665/commits/c65598a87ed172aed8c88df9b8832ae9070e8cff

dduportal:
The updatecli check error is the same as the `main` branch: unrelated to this PR as I'm not changing updatecli yet (despite the `1.0.0` tag)

lemeurherve (approved):

dduportal:
Putting back to draft until @lemeurherve published his patch on the 0.x line to unblokc him (and until updatecli is fixed)

dduportal:
Ready to roll

lemeurherve (approved):

PR #681, opened 2023-09-07
chore(mirrorbits-parent): ignore `charts` folder

lemeurherve:
Ref: #676

### DIFF
--- a/charts/mirrorbits-parent/.gitignore
+++ b/charts/mirrorbits-parent/.gitignore
@@ -1,0 +1,2 @@
+Chart.lock
+charts/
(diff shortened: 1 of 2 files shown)


charts/mirrorbits-parent/.gitignore
```diff
@@ -0,0 +1,2 @@
+Chart.lock
```
dduportal, line 1:
```suggestion
```

https://helm.sh/docs/helm/helm_dependency_build/#synopsis

> If no lock file is found, 'helm dependency build' will mirror the behavior of 'helm dependency update'

dduportal (approved):

PR #685, opened 2023-09-07
chore(updatecli/rsyncd): use `patch` as versionincrement

lemeurherve:
Like the other charts.

Noticed in #683


dduportal (approved):
Good catch

PR #710, opened 2023-09-15
feat(rsyncd): add the possibility to create PV and PVC

lemeurherve:


dduportal (changes requested):
I'm not in favor of this feature. Creating a dynamically provisioned PV+PVC is a use case we don't have: better to provision statically either in Terraform and/or as a feature of the `mirrorbits-parent` chart.

lemeurherve:
Closed in favor of https://github.com/jenkins-infra/helm-charts/pull/774

PR #712, opened 2023-09-18
chore(tests) remove all `test-connection.yaml` tests as useless

smerle33:
found out that multiple charts got their /tests folder misplaced, moved them to the root of each chart.


dduportal (approved):

hervelemeur:
> found out that multiple charts got their /tests folder misplaced, moved them to the root of each chart.

For the record, these folders inside "templates" folder correspond to the default e2e tests created via `helm create`, they were not misplaced and should not be confused with the helm-unittests we're adding in <chart>/tests

dduportal:
> > found out that multiple charts got their /tests folder misplaced, moved them to the root of each chart.
> 
> For the record, these folders inside "templates" folder correspond to the default e2e tests created via `helm create`, they were not misplaced and should not be confused with the helm-unittests we're adding in /tests

Yup: it is documented in https://helm.sh/docs/topics/chart_tests/.

We removed these tests since we do NOT either run or maintain those tests

Issue #655, opened 2023-09-06
Fix `acme` chart issue: `could not find expected ':'`


lemeurherve:
`helm template acme` output:

> Error: YAML parse error on acme/templates/clusterIssuer.yaml: error converting YAML to JSON: yaml: line 14: could not find expected ':'

Unit tests should be added, doc: https://github.com/helm-unittest/helm-unittest/blob/main/DOCUMENT.md

linoleparquet:
Hi, I created a PR to fix this one. Let me know what you think.

PR #715, opened 2023-09-19
fix(acme): fix templating and add unit tests (#655)

linoleparquet:
Related to #655 

Hi folks !
Fisrt PR on open source project. 
I renamed `acme_test_secret` to `acme-test-secret` because helm lint raised a warning.

### DIFF
--- a/charts/acme/tests/defaults_test.yaml
+++ b/charts/acme/tests/defaults_test.yaml
@@ -1,0 +1,23 @@
+suite: default tests
+values:
+  - ../values.yaml
+  - secrets_default_values.yaml
+templates:
+  - clusterIssuer.yaml
+  - secret.yaml
+tests:
+  - it: should create a cluster issuer by default
+    template: clusterIssuer.yaml
+    asserts:
+      - hasDocuments:
+          count: 1
+  - it: should create a Secret with the correct value
+    template: secret.yaml
+    asserts:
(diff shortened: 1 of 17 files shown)


charts/acme/tests/defaults_test.yaml
```diff
@@ -0,0 +1,23 @@
+suite: default tests
+values:
+  - ../values.yaml
+  - secrets_default_values.yaml
```
dduportal, line 4:
```suggestion
```

The "default test" should not provide any (helm) value at all as it is expected to simulate the `helm template ./` command.

I believe it *might* break the test and could require more changes (sorry 😅)

dduportal (changes requested):
Thanks for the proposed contribution!

It is in a good direction but require a bit more work (see my suggestion and comment) as we expect the default unit test suite to not provide any values

charts/acme/tests/defaults_test.yaml
linoleparquet: Got your point
I split the unit test in two files: custom test and default test. Let me know if that looks good to you.

I actually took example of the accountapp chart to redact theses tests, which doesn't follow this structure. If you create an issue I can fix it !

dduportal (approved):
LGTM, many thanks!

Issue #663, opened 2023-09-06
Fix `rss2twitter` chart issue: `nil pointer evaluating interface {}.consumerKey`


lemeurherve:
`helm template rss2twitter` output:

> Error: template: rss2twitter/templates/secret.yaml:9:30: executing "rss2twitter/templates/secret.yaml" at <.Values.twitter.consumerKey>: nil pointer evaluating interface {}.consumerKey

Unit tests should be added, doc: https://github.com/helm-unittest/helm-unittest/blob/main/DOCUMENT.md

dduportal:
- Release after #716 was merged
- Deployed to production in https://github.com/jenkins-infra/kubernetes-management/pull/4378

👏

PR #716, opened 2023-09-19
fix(rss2twitter): fix templating and add unit tests (#663)

linoleparquet:
Fix the helm chart templating issue. Added a basic README as well.

### DIFF
--- a/charts/rss2twitter/tests/defaults_test.yaml
+++ b/charts/rss2twitter/tests/defaults_test.yaml
@@ -1,0 +1,39 @@
+suite: default tests
+values:
+  - ../values.yaml
+  - secrets_default_values.yaml
+templates:
+  - _helper.tpl
+  - deployment.yaml
+  - exclusion-patterns-configmap.yaml
+  - secret.yaml
+tests:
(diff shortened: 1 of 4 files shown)


charts/rss2twitter/tests/defaults_test.yaml
```diff
@@ -0,0 +1,39 @@
+suite: default tests
+values:
+  - ../values.yaml
+  - secrets_default_values.yaml
```
dduportal, line 4:
```suggestion
```

Same as https://github.com/jenkins-infra/helm-charts/pull/715#discussion_r1331249598

dduportal (changes requested):
Same as https://github.com/jenkins-infra/helm-charts/pull/715#discussion_r1331249598: this is a really nice contribution and it is going in the right direction!

charts/rss2twitter/tests/defaults_test.yaml
linoleparquet: Done

dduportal (approved):
LGTM, that looks like a nice work 👏 

THanks!

PR #730, opened 2023-09-21
feat(mirrorbits) use a volume for logs

dduportal:
While working on https://github.com/jenkins-infra/helpdesk/issues/3136, I had to enable the download logs on mirrorbits.
These logs are written in the container root which is a bad practice (for performance at least).

This PR ensures that the logs directory of mirrorbits deployment is set as a data volume, emptydir by default.

- Added unit tests
- Installed on a local k3s (smoke test only: I haven't checked if the logs are writtenà)
- Tried a helmfile diff with production: it only changes the version in metadata and set the emptydir volume in the pod (no other changes)

(edit after first code review) This PR also updates the "custom values" test harness to factorize the values on top of the file. Not strictly mandatory but it helped me to properly set up my unit test case (as part of TDD steps).
Exception for the name overrides which are having a behavior we don't fully understand yet: subject of another PR.

### DIFF
--- a/charts/mirrorbits/tests/custom_values_test.yaml
+++ b/charts/mirrorbits/tests/custom_values_test.yaml
@@ -1,6 +1,28 @@
 suite: Tests with custom values
-values:
-  - secrets_default_values.yaml
+set:
+  fullNameOverride: "my-fullNameOverride"
+  ingress:
+    enabled: true
+    hosts:
+      - host: chart-example.local
+        paths:
+        - path: /
+          pathType: IfNotPresent
+  image:
+    files:
+      pullPolicy: Never
+    mirrorbits:
+      pullPolicy: Always
+  geoipupdate:
+    account_id: my-account-id
+    license_key: my-license-key
+  rsyncd:
+    enabled: true
+  mirrorbits:
+    logs:
+      volume:
+        persistentVolumeClaim:
+          claimName: foobar
 templates:
   - deployment.yaml
   - deployment.files.yaml
@@ -12,8 +34,6 @@ templates:
 tests:
   - it: Should set the correct service selector labels when a fullNameOverride is specified
     template: service.files.yaml
-    set:
-      fullNameOverride: "my-fullNameOverride"
     asserts:
       - hasDocuments:
           count: 1
@@ -28,6 +48,7 @@ tests:
   - it: Should set the correct service selector labels when a nameOverride is specified
     template: service.files.yaml
     set:
+      # Overides the fullname overide \o/
       nameOverride: "my-nameOverride"
     asserts:
       - hasDocuments:
@@ -42,10 +63,6 @@ tests:
           value: "RELEASE-NAME-files"
   - it: should define a customized "files" deployment
     template: deployment.files.yaml
-    set:
-      image:
-        files:
-          pullPolicy: Never
     asserts:
       - hasDocuments:
           count: 1
@@ -56,10 +73,6 @@ tests:
           value: Never
   - it: should define a customized "mirrorbits" deployment
     template: deployment.yaml
-    set:
-      image:
-        mirrorbits:
-          pullPolicy: Always
     asserts:
       - hasDocuments:
           count: 1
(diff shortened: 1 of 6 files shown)


charts/mirrorbits/tests/custom_values_test.yaml
```diff
@@ -28,6 +48,7 @@ tests:
   - it: Should set the correct service selector labels when a nameOverride is specified
     template: service.files.yaml
     set:
+      # Overides the fullname overide \o/
```
lemeurherve, line 51:
```suggestion
      # Overrides the name override \o/
```
nameOverride != fullnameOverride

charts/mirrorbits/tests/custom_values_test.yaml
```diff
@@ -12,8 +34,6 @@ templates:
 tests:
   - it: Should set the correct service selector labels when a fullNameOverride is specified
     template: service.files.yaml
-    set:
-      fullNameOverride: "my-fullNameOverride"
```
lemeurherve, line 16: I don't understand why this set has been removed, this is what this test is testing 🤔

charts/mirrorbits/tests/custom_values_test.yaml
```diff
@@ -42,10 +63,6 @@ tests:
           value: "RELEASE-NAME-files"
   - it: should define a customized "files" deployment
     template: deployment.files.yaml
-    set:
```
lemeurherve, line 45: Do these tests modifications need to be in this PR? If not, should be in a distinct one IMO.

charts/mirrorbits/tests/defaults_test.yaml
```diff
@@ -33,24 +30,7 @@ tests:
     asserts:
       - hasDocuments:
           count: 0
-  - it: should create ingress with pathType set to Prefix by default
```
lemeurherve, line 36: Same as above, mandatory in this PR?

charts/mirrorbits/tests/custom_values_test.yaml
dduportal: if both are specified, then `nameOveride` wins (hence the `fullnameOverride` set by default on the top level custom value)

charts/mirrorbits/tests/custom_values_test.yaml
dduportal: read the diff. it has been moved on top.

charts/mirrorbits/tests/custom_values_test.yaml
dduportal: Not completly, but the unit tests were spreading values and it started to be hard to know which suites in the custom values was handling which use case. hence the change to ensure my addons were coherent.

charts/mirrorbits/tests/defaults_test.yaml
dduportal: not mandatory but clearly useful

charts/mirrorbits/tests/custom_values_test.yaml
dduportal: Changed as per your review feedbacks

charts/mirrorbits/tests/custom_values_test.yaml
dduportal: Changed as per your review feedbacks

charts/mirrorbits/tests/defaults_test.yaml
```diff
@@ -76,19 +56,35 @@ tests:
       - equal:
           path: spec.template.metadata.labels["app.kubernetes.io/instance"]
           value: "RELEASE-NAME-files"
-  - it: should define the default "files" deployment with default imagePullPolicy and metadata labels
-    template: deployment.files.yaml
```
lemeurherve, line 80: No more test on this specific deployment?

charts/mirrorbits/values.yaml
```diff
@@ -86,7 +86,13 @@ geoipupdate:
   editions: GeoLite2-ASN GeoLite2-City GeoLite2-Country
   update_frequency: 24
 mirrorbits:
-  # mirrorbits.conf data, to be completed as secret with Redis credentials
+  logs:
+    volume:
+      emptyDir: {}
+      ## Volume contains the Pod Volume definition. For a PVC:
```
dduportal, line 92:
```suggestion
      ## Volume contains the Pod Volume definition. For an existing PVC for instance:
```

charts/mirrorbits/values.yaml
```diff
@@ -86,7 +86,13 @@ geoipupdate:
   editions: GeoLite2-ASN GeoLite2-City GeoLite2-Country
   update_frequency: 24
 mirrorbits:
-  # mirrorbits.conf data, to be completed as secret with Redis credentials
+  logs:
+    volume:
+      emptyDir: {}
+      ## Volume contains the Pod Volume definition. For an existing PVC for instance:
```
dduportal, line 92:
```suggestion
      ## Volume contains the Pod Volume definition. Example with an existing PVC:
```

PR #731, opened 2023-09-21
feat(mirrorbits-lite) use a volume for logs

dduportal:
Cherry pick of #730 and #732 to `mirrorbits-lite`

Includes an update of the chart metadatas

### DIFF
--- a/charts/mirrorbits-lite/Chart.yaml
+++ b/charts/mirrorbits-lite/Chart.yaml
@@ -1,4 +1,7 @@
-apiVersion: v1
+appVersion: v0.5.2
 description: Mirrobits lite helm chart for Kubernetes
 name: mirrorbits-lite
-version: 0.0.2
+version: 0.1.0
+maintainers:
+- email: jenkins-infra-team@googlegroups.com
+  name: jenkins-infra-team
--- a/charts/mirrorbits-lite/Chart.yaml
+++ b/charts/mirrorbits-lite/Chart.yaml
@@ -1,7 +1,9 @@
+apiVersion: v1
 appVersion: v0.5.2
 description: Mirrobits lite helm chart for Kubernetes
 name: mirrorbits-lite
 version: 0.1.0
+appVersion: v0.5.2
 maintainers:
 - email: jenkins-infra-team@googlegroups.com
   name: jenkins-infra-team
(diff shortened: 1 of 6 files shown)


charts/mirrorbits-lite/Chart.yaml
```diff
@@ -1,4 +1,7 @@
-apiVersion: v1
```
lemeurherve, line 1: apiVersion is mandatory AFAIK

lemeurherve (approved):

charts/mirrorbits-lite/Chart.yaml
```diff
@@ -1,4 +1,7 @@
...
+appVersion: v0.5.2
```
lemeurherve, line 1:
```suggestion
apiVersion: v1
appVersion: v0.5.2
```

lemeurherve:
```suggestion
```
And appVersion isn't (nit)

dduportal: Good catch for the confusion I had between `appVersion` and `apiVersion` Thanks!

`appVersion` indicate the version of the application provided, which is (and should) always be different than the version of the Docker image. In our case it is definitively different as the version schem of jenkinsciinfra/mirrorbits is different than mirrorbits version scheme

dduportal: Fixed by https://github.com/jenkins-infra/helm-charts/pull/731/commits/d9f4998bc4b6485670cf494d39d4c4cc4b5c793b

charts/mirrorbits-lite/Chart.yaml
```diff
@@ -1,4 +1,8 @@
 apiVersion: v1
 description: Mirrobits lite helm chart for Kubernetes
 name: mirrorbits-lite
-version: 0.0.2
+version: 0.1.0
+appVersion: v0.5.2
```
lemeurherve, line 5:
```suggestion
```
I still think there is zero added value with this (nit)

lemeurherve (approved):
LGTM

lemeurherve (approved):

charts/mirrorbits-lite/Chart.yaml
dduportal: It allows readers to immediately understand that the image tag version scheme is different than the application.
It's recommended by Helm doc but optional and it does not change the behavior (and we have it on the former `mirrorbits` chart?

PR #736, opened 2023-09-22
fix(mirrorbits) remove unneeded service account

dduportal:
Kubernetes Service Accounts are only required to interact with a Kubernetes API.

The mirrorbits service (and the associated httpd and rsync components) have no need to access it.

This PR removes any reference to the service account as it was created by the default helm template.

💡 Please note:

- A subsequent PR will remove the service account from the template to avoid propagating this "insane" default
- This PR needs a "twin" for `mirrorbits-lite`
- One PR per chart is required to remove SVC accounts


dduportal:
Superseeded by #737

PR #738, opened 2023-09-23
feat(mirrorbits-parent) add support for an umbrella ingress

dduportal:
This PR introduces the definition of the "umbrella" ingress we uses for get.jenkins.io and updates.jenkins.io as part of the mirrorbits-parent chart.

The rationale is that this ingress should redirect some requests to the `mirrorbits-lite` service (sub chart component) or to the `httpd` service (sub chart component): it cannot be defined by one or the other.

💡 Please note:

- I've redefined the attribute allowing to specify which backend to use as `backendService` which can take only 2 values: `httpd` or `mirrorbits`
- The usage of a carret (`-`) in the chart names is annoying when dealing with the values as the helm convention recommends not to use them. We had to use a different name than `mirrorbits` for the subchart so no easy way to change this. Hence I've used the `{{ index ... }}` trick to retrieve sub keys (see. for details https://github.com/helm/helm/issues/2192)
- The parent chart cannot, for today, automatically retrieve the subcharts service names as it is defined by a template.
  - We can use the `fullNameoverride` on the subcharts but the value is static (e.g. not templated) so we cannot install the same chart multiple time on the same namespace. 
  - We could use global value: I've added a TODO in comment.


charts/mirrorbits-parent/templates/ingress.yaml
```diff
@@ -0,0 +1,59 @@
...
+          - path: {{ .path }}
+            pathType: {{ .pathType | default "Prefix" }}
+            backend:
+            {{- if eq .backendService "mirrorbits" }}
+              {{- if not (index $.Values "mirrorbits-lite" "enabled") }}
+                {{- fail "Cannot use mirrorbits(-lite) as backend if it is disabled." }}
+              {{- end }}
+              service:
+                # TODO: update child chart to use a global value to let the parent-chart defined/get the value
+                name: {{ $.Release.Name | trunc 63 | trimSuffix "-" -}}-mirrorbits-lite
```
lemeurherve, line 42: All these names are risky: let say someone puts a really long release name like [Rindfleischetikettierungsuberwachungsaufgabenubertragungsgesetz](https://fr.wikipedia.org/wiki/Rindfleisch%C2%ADetikettierungs%C2%AD%C3%BCberwachungs%C2%ADaufgaben%C2%AD%C3%BCbertragungs%C2%ADgesetz), what happens then? (Spoiler: it will fail)

dduportal: You are correct. The naming here only relies on the subchart templating which is responsible for the size of the generated name through their template function.

I've deferred the resolution of this challenge as per the TODO comment. If we can pass an overiding name (even a long one) to the child chart, there should be no problem at all as the subchart templating function would truncate it (you already added smart logic in it already to deduplicate the release name btw.)

But it's not the role of the parent chart as per the encapsulation principle.

charts/mirrorbits-parent/templates/ingress.yaml
```diff
@@ -0,0 +1,59 @@
...
+                port:
+                  number: {{ index $.Values "mirrorbits-lite" "service" "port" }}
+            {{- else if eq .backendService "httpd" }}
+              {{- if not $.Values.httpd.enabled }}
```
lemeurherve, line 46: I'm really not fond of this `index` stuff required by the dash in the subchart name.

Looking forward to mirrorbits (old) chart deprecation so we can rename mirrorbits-lite and get rid of this incongruity.

lemeurherve (approved):

Issue #662, opened 2023-09-06
Fix `reports` chart issue: `could not find expected ':'`


lemeurherve:
`helm template reports` output:

> Error: YAML parse error on reports/templates/deployment.yaml: error converting YAML to JSON: yaml: line 62: could not find expected ':'

Unit tests should be added, doc: https://github.com/helm-unittest/helm-unittest/blob/main/DOCUMENT.md

linoleparquet:
Resolved in #772

dduportal:
@linoleparquet for information, issues can be closed automatically if there is a commit message (or in the PR body) with the correct textual form: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

linoleparquet:
Thanks for the tip !!

PR #772, opened 2023-09-23
fix(reports): fix helm template and add unit tests

linoleparquet:
Resolve Issue #662


charts/reports/templates/deployment.yaml
```diff
@@ -66,8 +66,10 @@ spec:
...
-        - name: config
-          configMap:
-            name: reports
+      {{- with .Values.htmlVolume  }}
+      - name: html
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      - name: config
+        configMap:
+          name: {{ include "reports.fullname" . }}
```
dduportal, line 75:
```suggestion
        {{- with .Values.htmlVolume  }}
        - name: html
          {{- toYaml . | nindent 8 }}
        {{- end }}
        - name: config
          configMap:
            name: {{ include "reports.fullname" . }}
```

nit: trying to diff this PR with an installed `reports` chart shows there is a minor indentation difference.

Non blocking as per https://yaml.org/spec/1.2.2/#61-indentation-spaces (both indent and non indent are OK) but I prefer to avoid bad surprises if you don't mind?

dduportal (commented):
Good job! Thanks for the contribution!

Only 1 blocker: you have to update the chart version number (patch bump at first sight) to have the checks passing

charts/reports/templates/deployment.yaml
linoleparquet: LGTM :+1: 
Commited the changes

dduportal (dismissed):
Nice job! I'm self-commiting a nit change (adding an endline to files is recommended for POSIX: https://www.baeldung.com/linux/files-end-with-newlines#newlines-at-the-end-of-the-file)

charts/reports/templates/deployment.yaml
```diff
@@ -66,8 +66,10 @@ spec:
...
       volumes:
+        {{- with .Values.htmlVolume  }}
         - name: html
-{{ toYaml .Values.htmlVolume | indent 10 }}
+          {{- toYaml . | nindent 10 }}
+        {{- end }}
         - name: config
           configMap:
-            name: reports
+            name: {{ include "reports.fullname" . }}
```
dduportal, line 75:
```suggestion
            name: {{ include "reports.fullname" . }}

```

dduportal, line 75: Good catch 👏

dduportal (approved):
LGTM!

- Tests added 🥳 
- 2 bug corrected in this PR 👏 
- Helmfile diff on production looks good!

Thanks!

charts/reports/templates/deployment.yaml
linoleparquet: Thanks !

PR #774, opened 2023-09-24
feat(mirrorbits-parent) allow managing an 'umbrella' persistent storage

dduportal:
This PR is a first step towards a full integration of the persistent storage management and the sub-components.

It allows managing the PV and PVC from the parent chart with the added value of synchronizing properly the properties between both.

It's the same idea as #710 but, like #738 , factorized to the parent chart so the parent chart manages the shared storage.

It does not have ... yet... an automated value passing to the sub charts: subsequent PR. It means for now, we have to give the pvc reference to the subcharts when installing it.

💡 Notes:

- I've split some the test harness per resource to avoid big and long files AND because the `set` directive are cumulatives AND sequential (meaning the order of sub-test harnesses matters, leading to weird behavior).


dduportal:
Putting back as draft after the first real life tentative as it is missing some elements (secret to allow accessing the azurefile in production)

dduportal (commented):
This updated charts can be deployed with success on my production test (not only unit tests):

```
$ kubectl get pvc,pod,svc,ingress,secret -n get-jenkins-io
NAME                                                     STATUS   VOLUME                             CAPACITY   ACCESS MODES   STORAGECLASS            AGE
persistentvolumeclaim/get-jenkins-io-mirrorbits-parent   Bound    get-jenkins-io-mirrorbits-parent   1000Gi     RWX            azurefile-csi-premium   5m55s

NAME                                                 READY   STATUS    RESTARTS   AGE
pod/get-jenkins-io-httpd-84b8d4b658-lpmq4            1/1     Running   0          5m55s
pod/get-jenkins-io-mirrorbits-lite-654c5d649-6fxjj   2/2     Running   0          5m55s

NAME                                     TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
service/get-jenkins-io-httpd             ClusterIP   10.0.240.131   <none>        80/TCP    5m55s
service/get-jenkins-io-mirrorbits-lite   ClusterIP   10.0.233.167   <none>        80/TCP    5m55s

NAME                                                         CLASS          HOSTS                                                                                    ADDRESS       PORTS     AGE
ingress.networking.k8s.io/get-jenkins-io-mirrorbits-parent   public-nginx   <redacted> + 1 more...   <redacted>   80, 443   5m56s

NAME                                                              TYPE                 DATA   AGE
secret/get-jenkins-io-mirrorbits-lite                             Opaque               1      5m56s
secret/get-jenkins-io-mirrorbits-lite-geoipupdate                 Opaque               2      5m56s
secret/get-jenkins-io-mirrorbits-parent-persistentvolume-secret   Opaque               2      5m56s
secret/mirrorbits-tls                                             kubernetes.io/tls    2      117m
```

I see a possible naming improvement for this PR: the persistentvolume secret name should be templatized and truncated to avoid too long name.

dduportal:
Added https://github.com/jenkins-infra/helm-charts/pull/774/commits/5ee1f795800c25e730312ce87bd8c804ccb3d67b to handle long name. Looks good for this first iteration.

charts/mirrorbits-parent/values.yaml
```diff
@@ -27,6 +27,44 @@ ingress:
   #     hosts:
   #       - company.org
 
+## Persistent Storage used across components
+storage:
+  ## Manage (and reuse) PV + PVC for all components
+  manage: false
```
timja, line 33: would `enabled` be better? more consistent with how enabling a feature normally.

timja (approved):

charts/mirrorbits-parent/values.yaml
dduportal: Fixed by https://github.com/jenkins-infra/helm-charts/pull/774/commits/d38783bdca48a5734850903e119f9e1e6d46d3e8

timja (approved):

charts/mirrorbits-parent/templates/_helpers.tpl
```diff
@@ -30,3 +30,10 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end -}}
+
+{{/*
+Create chart name and version as used by the chart label.
```
lemeurherve, line 35: This comment should be adapted (copied from L15 I suppose)

charts/mirrorbits-parent/templates/persistentvolume.yaml
```diff
@@ -0,0 +1,30 @@
...
+{{- include "mirrorbits-parents.labels" . | nindent 4 }}
+spec:
+  accessModes:
+    {{- toYaml .Values.storage.accessModes | nindent 4 }}
+  capacity:
+    storage: {{ .Values.storage.storageSize }}
+  storageClassName: {{ .Values.storage.storageClassName }}
+  {{- with .Values.storage.persistentVolume.azureFile }}
+  csi:
+    driver: file.csi.azure.com
```
lemeurherve, line 17: This mandates the deployment on Azure then, should a note be added in the README or somewhere?

I suppose keeping a more universal structure would be too restrictive?

lemeurherve:
Erf, I was adding review comments...

charts/mirrorbits-parent/templates/persistentvolumesecret.yaml
```diff
@@ -0,0 +1,13 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "mirrorbits-parent.pv-secretname" . }}
+  labels:
+{{ include "mirrorbits-parents.labels" . | nindent 4 }}
+type: Opaque
+data:
+  azurestorageaccountname: {{ .Values.storage.persistentVolume.azureFile.storageAccountName | b64enc }}
+  azurestorageaccountkey: {{ .Values.storage.persistentVolume.azureFile.storageAccountKey | b64enc }}
```
lemeurherve, line 12: No camelCase for the data field names?

lemeurherve (commented):
For posterity 🤷

charts/mirrorbits-parent/templates/persistentvolume.yaml
dduportal: Only if you specify the `storage.persistentVolume.azureFile` value (empty by default): you can use this on DigitalOcean (or with an Azure disk) by using the `storage.persistentVolume.additionalSpec` for instance.

However you are correct, a bit of documentation is needed.

charts/mirrorbits-parent/templates/persistentvolumesecret.yaml
dduportal: https://learn.microsoft.com/en-us/azure/aks/azure-csi-files-storage-provision#create-a-kubernetes-secret : nope (and we don't have for the current production secrets either).

As much as CamelCase convention makes sens for values file, in templates we must stick to either Kubernetes or Cloud providers syntax.

dduportal:
> Erf, I was adding review comments...

I'm opening a new PR with docs and feedback. Shoot.

charts/mirrorbits-parent/templates/persistentvolume.yaml
dduportal: https://github.com/jenkins-infra/helm-charts/pull/775

charts/mirrorbits-parent/templates/_helpers.tpl
dduportal: https://github.com/jenkins-infra/helm-charts/pull/775

dduportal:
> > Erf, I was adding review comments...
> 
> I'm opening a new PR with docs and feedback. Shoot.

https://github.com/jenkins-infra/helm-charts/pull/775 go go go

PR #779, opened 2023-09-25
docs(mirrorbits,mirrorbits-lite) move Jenkins production custom documentation to only runbooks

dduportal:
While working on https://github.com/jenkins-infra/helpdesk/issues/3136, I realized that we duplicate a lot of production-related information while it should be on a single location: (private runbook) https://github.com/jenkins-infra/runbooks/tree/main/mirrorbits


dduportal:
Self merging: doc only

PR #780, opened 2023-09-25
fix(account-app) correct default values for 'resources' to avoid YAML parsing error

dduportal:
Thanks @olblak for catching the root issue ❤️ 

This PR fixes an issue when using the default value for `resources` which should be `{}` to ensure all YAML parser consider it as null.

Please note that I have a personal preference for commenting out the empty or undefined values on helm charts, but this fix will do.

### DIFF
--- a/charts/accountapp/values.yaml
+++ b/charts/accountapp/values.yaml
@@ -29,8 +29,8 @@ ingress:
 #    - secretName: accounts-tls
 #      hosts:
 #        - accounts.jenkins.io
-resources: # We usually recommend not to specify default resources and to leave this as a conscious
-# choice for the user. This also increases chances charts run on environments with little
+# We usually recommend not to specify default resources and to leave this as a conscious: null # choice for the user. This also increases chances charts run on environments with little
+resources: {}
 # resources, such as Minikube. If you do want to specify resources, uncomment the following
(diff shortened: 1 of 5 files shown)


dduportal:
- Added unit tests
- An helmfile `diff` on the production shows no change (except version meta)

dduportal:
Self-merging as this PR fixes the updatecli build

charts/accountapp/values.yaml
```diff
@@ -29,8 +29,8 @@ ingress:
 #    - secretName: accounts-tls
 #      hosts:
 #        - accounts.jenkins.io
-resources: # We usually recommend not to specify default resources and to leave this as a conscious
-# choice for the user. This also increases chances charts run on environments with little
+# We usually recommend not to specify default resources and to leave this as a conscious: null # choice for the user. This also increases chances charts run on environments with little
```
hervelemeur, line 32: `null` or empty object? The comment seems to imply one should change the defaut `{}` value to `null`

Also, should this be splited to two lines?

dduportal: Good catch. The fix with `with` helm control flow element does not care as the resulting template pipeline is `false` wether the value is null, empty or undefined. 

PR to update the comment?

PR #815, opened 2023-10-02
feat!(mirrorbits) 💥 allow specifying custom affinities per deployment

smerle33:
need to update https://github.com/jenkins-infra/kubernetes-management/pull/4431/

### DIFF
--- a/charts/mirrorbits/tests/custom_values_test.yaml
+++ b/charts/mirrorbits/tests/custom_values_test.yaml
@@ -22,6 +22,46 @@ set:
       volume:
         persistentVolumeClaim:
           claimName: foobar
+  affinity:
+    mirrorbits:
+      podAntiAffinity:
+        requiredDuringSchedulingIgnoredDuringExecution:
+          - labelSelector:
+              matchExpressions:
+                - key: "app.kubernetes.io/name"
+                  operator: In
+                  values:
+                    - mirrorbits
+            topologyKey: "kubernetes.io/hostname"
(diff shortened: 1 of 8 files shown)


charts/mirrorbits/tests/custom_values_test.yaml
```diff
@@ -94,6 +134,9 @@ tests:
       - equal:
           path: spec.template.spec.containers[0].volumeMounts[2].mountPath
           value: /var/log/mirrorbits
+      - equal:
+          path: spec.template.spec.affinity.mirrorbits.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].topologyKey
```
dduportal, line 138:
```suggestion
          path: spec.template.spec.affinity.podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecution[0].topologyKey
```

Otherwise the semantic is invalid (Kubernetes does not allow the `mirrorbits` keyword as child of `affinity`).

dduportal (changes requested):
You have to:

- Use the proper test (remove the "component" name aka. `mirrorbits`) => will make the test fail as expected because there is no implementation yet
- Test the 3 deployments (mirrorbits, httpd and rsyncd)
- Update the template of the 3 deployments to insert the affinity.<component name>.<value> in the `affinity` field


=> you can refer to what we did on jenkinsio's affinities

dduportal (approved):

PR #817, opened 2023-10-02
feat(mirrorbits-lite) allow defining ingress and PV(C)s from parent chart

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/2649

This PR is an extract of https://github.com/jenkins-infra/helm-charts/pull/813 scoped to only `mirrorbits-lite`

chore(mirrorbits-lite) increase unit test coverage to PV and PVC with their volumeMounts

PR #818, opened 2023-10-03
feat(httpd) allow overriding ingress and PV(C)s based on global value from parent chart

dduportal:
Same as https://github.com/jenkins-infra/helm-charts/pull/817 but for httpd

### DIFF
--- a/charts/httpd/tests/defaults_test.yaml
+++ b/charts/httpd/tests/defaults_test.yaml
@@ -3,7 +3,7 @@ templates:
   - configmap.yaml
   - deployment.yaml
   - ingress.yaml
-  - secret.yaml # Direct dependency of deployment(.*).yaml
+  - secret.yaml # Direct dependency of deployment.yaml
   - service.yaml
   - persistentVolume.yaml
   - persistentVolumeClaim.yaml
--- a/charts/httpd/tests/defaults_test.yaml
+++ b/charts/httpd/tests/defaults_test.yaml
@@ -5,30 +5,15 @@ templates:
   - ingress.yaml
   - secret.yaml # Direct dependency of deployment(.*).yaml
   - service.yaml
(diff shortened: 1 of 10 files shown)


charts/httpd/tests/defaults_test.yaml
```diff
@@ -5,30 +5,15 @@ templates:
   - ingress.yaml
   - secret.yaml # Direct dependency of deployment(.*).yaml
```
lemeurherve, line 6:
```suggestion
  - secret.yaml # Direct dependency of deployment.yaml
```
(nit; remnant of mirrorbits)

charts/httpd/tests/defaults_test.yaml
```diff
@@ -68,17 +40,40 @@ tests:
           path: spec.template.metadata.labels["app.kubernetes.io/instance"]
           value: "RELEASE-NAME"
       - equal:
-          path: spec.template.spec.containers[0].imagePullPolicy
+          path: "spec.template.spec.containers[*].imagePullPolicy"
```
lemeurherve, line 43: Can there be more than one container? If not, I would have kept `0` to be like the other tests below. (nit)

lemeurherve (approved):

lemeurherve:
Note: I've edited the PR title.

charts/httpd/tests/defaults_test.yaml
dduportal: Yes there can be multiple containers (`spec.template.spec.containers` is a list).
The test here checks the imagepullpolicy set for any containers, it's the goal.

The test below have a different meaning and only checks the first container: it's not a matter of codestyle of homogeneity, but a matter of intent.

lemeurherve (approved):

PR #820, opened 2023-10-03
feat(reports): add PodDisruptionBudget and unittests

smerle33:

### DIFF
--- a/charts/reports/templates/_helpers.tpl
+++ b/charts/reports/templates/_helpers.tpl
@@ -32,12 +32,19 @@ Create chart name and version as used by the chart label.
 {{- end -}}
 
 {{/*
+Selector labels
+*/}}
+{{- define "reports.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "reports.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+{{/*
 Common labels
 */}}
 {{- define "reports.labels" -}}
-app.kubernetes.io/name: {{ include "reports.name" . }}
+{{ include "reports.selectorLabels" . }}
 helm.sh/chart: {{ include "reports.chart" . }}
-app.kubernetes.io/instance: {{ .Release.Name }}
 {{- if .Chart.AppVersion }}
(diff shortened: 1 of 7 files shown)


charts/reports/templates/_helpers.tpl
```diff
@@ -31,13 +31,20 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{/*
+Selector labels
+*/}}
+{{- define "reports.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "reports.name" . }}
```
dduportal, line 38: Don't forget the unit test for this change (check the `metadata.labels` for non regression ;) )

charts/reports/templates/deployment.yaml
```diff
@@ -51,8 +51,10 @@ spec:
               mountPath: /usr/share/nginx/html
             - name: config
               mountPath: /etc/nginx/conf.d
+          {{- with .Values.resources }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
+          {{- end}}
```
dduportal, line 57:
```suggestion
          {{- with .Values.resources }}
          resources:
            {{- toYaml . | nindent 12 }}
          {{- end}}
```

Otherwise:

```
Error: Failed to render chart: exit status 1: Error: template: reports/templates/deployment.yaml:56:30: executing "reports/templates/deployment.yaml" at <.Values.resources>: nil pointer evaluating interface {}.resources
```

(worth a unit test with custom values to catch this error, then fix it with my suggestion)

dduportal (changes requested):
Error when running on production:

```
Error: Failed to render chart: exit status 1: Error: template: reports/templates/deployment.yaml:56:30: executing "reports/templates/deployment.yaml" at <.Values.resources>: nil pointer evaluating interface {}.resources
```

see my suggestion (otherwise nice work!)

dduportal (approved):

PR #821, opened 2023-10-03
feat(uplink): add PodDisruptionBudget and unittests

smerle33:

### DIFF
--- a/charts/uplink/templates/_helpers.tpl
+++ b/charts/uplink/templates/_helpers.tpl
@@ -32,12 +32,20 @@ Create chart name and version as used by the chart label.
 {{- end -}}
 
 {{/*
+Selector labels
+*/}}
+{{- define "uplink.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "uplink.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end }}
+
+
+{{/*
 Common labels
 */}}
 {{- define "uplink.labels" -}}
-app.kubernetes.io/name: {{ include "uplink.name" . }}
+{{ include "uplink.selectorLabels" . }}
 helm.sh/chart: {{ include "uplink.chart" . }}
-app.kubernetes.io/instance: {{ .Release.Name }}
 {{- if .Chart.AppVersion }}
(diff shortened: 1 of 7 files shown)


charts/uplink/templates/_helpers.tpl
```diff
@@ -31,13 +31,21 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{/*
+Selector labels
+*/}}
+{{- define "uplink.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "uplink.name" . }}
```
dduportal, line 38: Don't forget the unit test (on the deployment) for this change ;)

charts/uplink/templates/pdb.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if (gt (int .Values.replicaCount) 1) }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ include "uplink.fullname" . }}
+  labels:
+    {{- include "uplink.labels" . | nindent 4 }}
+spec:
+  {{- with .Values.poddisruptionbudget.minAvailable }}
```
dduportal, line 9: `.Values.poddisruptionbudget` is undefined causing the error `Error: Failed to render chart: exit status 1: Error: template: uplink/templates/pdb.yaml:9:18: executing "uplink/templates/pdb.yaml" at <.Values.poddisruptionbudget.minAvailable>: nil pointer evaluating interface {}.minAvailable`.

You might want to look at  https://github.com/jenkins-infra/helm-charts/pull/822/files and add the corresponding tests (first) and fix

charts/uplink/templates/_helpers.tpl
smerle33: done with https://github.com/jenkins-infra/helm-charts/pull/821/commits/003a84b7e106012ca1cf83255e858d72d772101c

dduportal (changes requested):
Almost there! There is still one issue in the PDB though (see my comment)

PR #822, opened 2023-10-03
fix(artifact-caching-proxy) set defaults on PDB

dduportal:
This PR fixes the error seen in https://github.com/jenkins-infra/kubernetes-management/pull/4460#pullrequestreview-1649313575 

```
[ERROR] templates/: template: artifact-caching-proxy/templates/pdb.yaml:9:18: executing "artifact-caching-proxy/templates/pdb.yaml" at <.Values.poddisruptionbudget.minAvailable>: nil pointer evaluating interface {}.minAvailable
```

💡Notes:

- Unit test added to cover the case when a PDB is created because multiple replicas are set up but with its default setup. This unit tests returns the error above with the version `0.15.0` of this chart
- Unit test added to verify the coverage of custom value `maxUnavailable`

### DIFF
--- a/charts/artifact-caching-proxy/Chart.yaml
+++ b/charts/artifact-caching-proxy/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 description: artifact-caching-proxy is a Nginx caching proxy in front of repo.jenkins-ci.org
 name: artifact-caching-proxy
 type: application
-version: 0.15.0
+version: 0.15.1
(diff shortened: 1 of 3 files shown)


dduportal:
Info: A diff on production with this fix shows the following:

```
+ # Source: artifact-caching-proxy/templates/pdb.yaml
+ apiVersion: policy/v1
+ kind: PodDisruptionBudget
+ metadata:
+   name: artifact-caching-proxy
+   labels:
+     helm.sh/chart: artifact-caching-proxy-0.15.1
+     app.kubernetes.io/name: artifact-caching-proxy
+     app.kubernetes.io/instance: artifact-caching-proxy
+     app.kubernetes.io/managed-by: Helm
+ spec:
+   minAvailable: 1
+   selector:
+     matchLabels:
+       app.kubernetes.io/name: artifact-caching-proxy
+       app.kubernetes.io/instance: artifact-caching-proxy
```

which looks good to me!

lemeurherve (approved):

PR #836, opened 2023-10-06
fix(mirrorbit-lite) only create secrets when needed

dduportal:
While working on the parent chart for https://github.com/jenkins-infra/helpdesk/issues/2649, caught this tiny bug in the helm chart where secrets where created even if values are not provided.

This PR fixes the bug by adding conditions to create the secrets (and tests).

### DIFF
--- a/charts/mirrorbits-lite/tests/defaults_test.yaml
+++ b/charts/mirrorbits-lite/tests/defaults_test.yaml
@@ -73,7 +73,7 @@ tests:
     asserts:
       - hasDocuments:
           count: 0
-  - it: should not create only create the mirrorbits conf secret by default
+  - it: should only create the mirrorbits conf secret by default
     template: secrets.yaml
     asserts:
       - hasDocuments:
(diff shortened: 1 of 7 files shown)


charts/mirrorbits-lite/tests/defaults_test.yaml
```diff
@@ -64,3 +64,13 @@ tests:
     asserts:
       - hasDocuments:
           count: 0
+  - it: should not create only create the mirrorbits conf secret by default
```
lemeurherve, line 67:
```suggestion
  - it: should only create the mirrorbits conf secret by default
```

charts/mirrorbits-lite/tests/parent_values_test.yaml
```diff
@@ -91,3 +96,13 @@ tests:
     asserts:
       - hasDocuments:
           count: 0
+  - it: should not create only create the mirrorbits conf secret (global storage overrides local PVC secrets)
```
lemeurherve, line 99:
```suggestion
  - it: should only create the mirrorbits conf secret (global storage overrides local PVC secrets)
```

lemeurherve (approved):

lemeurherve (approved):

PR #843, opened 2023-10-06
fix(plugin-site) correct labels of the frontend deployment

dduportal:

PR #844, opened 2023-10-06
hotfix(plugin-site) separate svcs

dduportal:
hotfix tested in production. Fixes up #843 and the chart by splitting into 2 services and correcting the ingress warnings.

Avoids a LOT of dropped connection internally to the cluster


dduportal:
Self merging

PR #848, opened 2023-10-09
feat(plugin-site) bump backend image tag to `294-5389ef`

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/3778

This PR embeds a few changes so we will have to watch carefully when deployed to production!

PR #850, opened 2023-10-10
chore(updatecli): add manifest for `plugin-site-api` docker image

lemeurherve:
Related: #848


updatecli/updatecli.d/plugin-site-api
```diff
@@ -0,0 +1,53 @@
...
+
+sources:
+  latestRelease:
+    kind: githubrelease
+    name: "Get latest jenkins-infra/plugin-site-api latest version"
+    spec:
+      owner: "jenkins-infra"
+      repository: "docker-plugin-site-issues"
+      token: "{{ requiredEnv .github.token }}"
+      username: "{{ .github.username }}"
```
dduportal, line 23: We haven't used (git) tags or (GitHub) releases since 2021 for this repository: https://github.com/jenkins-infra/plugin-site-api/tags

dduportal (changes requested):
This PR does not track the expected version. It looks like a source of type `dockerImage` should be used instead (and no conditions)

PR #871, opened 2023-10-17
Fix `rating` chart render issue and add unit tests

### LINKED ISSUE #661: Fix `rating` chart issue: `mapping values are not allowed in this context`
`helm template rating` output:

> Error: YAML parse error on rating/templates/deployment.yaml: error converting YAML to JSON: yaml: line 37: mapping values are not allowed in this context

Unit tests should be added, doc: https://github.com/helm-unittest/helm-unittest/blob/main/DOCUMENT.md

limejuny:
### Detail
 - Fix #661
 - `{{- if ~~ -}}` will be render inappropriately (ref: https://helm.sh/docs/chart_template_guide/control_structures/#controlling-whitespace)
   - remove trailing `-`, so it renders well.
 - Add simple unit test to ensure rendering.

### Testing
 - original `deployment.yaml`
   - image
 - fixed version
   - image

### DIFF
--- a/charts/rating/Chart.yaml
+++ b/charts/rating/Chart.yaml
@@ -3,4 +3,4 @@ appVersion: "1.0"
 description: A Helm chart for Kubernetes
 name: rating
 type: application
-version: 0.3.5
+version: 0.3.6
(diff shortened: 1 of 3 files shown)


dduportal (approved):
Good work, thanks for the contribution!

For info, I pushed an additional commit to bump the chart version to validate the check, hope you don't mind.

limejuny:
> Good work, thanks for the contribution!
> 
> For info, I pushed an additional commit to bump the chart version to validate the check, hope you don't mind.

Thanks!

PR #903, opened 2023-10-30
fix(uplink): use tag instead of sha256 for the Docker image

lemeurherve:
Using a sha256 blocks the deployment of the arm64 image by the chart, thus this PR switch to the named tag instead of the sha256 in the values.

Related:
- https://github.com/jenkins-infra/kubernetes-management/pull/4610

Note: the image specs override in kubernetes-management has been removed:
- https://github.com/jenkins-infra/kubernetes-management/pull/4607

### DIFF
--- a/charts/uplink/values.yaml
+++ b/charts/uplink/values.yaml
@@ -4,7 +4,7 @@
 replicaCount: 1
 image:
   repository: jenkinsciinfra/uplink@sha256
-  tag: e57f1ee08d37c985388ebc87aafd3e7a73ad0a1a950abc1d2eb00c093eb3e76b
+  tag: 0.0.2@sha256:e57f1ee08d37c985388ebc87aafd3e7a73ad0a1a950abc1d2eb00c093eb3e76b
   pullPolicy: IfNotPresent
 imagePullSecrets: []
 nameOverride: ""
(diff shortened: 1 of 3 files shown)


updatecli/updatecli.d/uplink.yaml
```diff
@@ -20,7 +20,6 @@ sources:
       image: "jenkinsciinfra/uplink"
       tag: "latest"
       architecture: "amd64"
-      hidetag: true
     transformers:
       - trimprefix: '@sha256:'
```
lemeurherve, line 24:
```suggestion
```

charts/uplink/values.yaml
```diff
@@ -4,7 +4,7 @@
 replicaCount: 1
 image:
   repository: jenkinsciinfra/uplink@sha256
-  tag: e57f1ee08d37c985388ebc87aafd3e7a73ad0a1a950abc1d2eb00c093eb3e76b
+  tag: 0.0.2@sha256:e57f1ee08d37c985388ebc87aafd3e7a73ad0a1a950abc1d2eb00c093eb3e76b
```
dduportal, line 7: That will fail, resulting in the image `jenkinsciinfra/uplink@sha256:0.0.2@sha256:e57f1ee08d37c985388ebc87aafd3e7a73ad0a1a950abc1d2eb00c093eb3e76b` being checked (which is invalid).

dduportal (changes requested):

dduportal (approved):

PR #911, opened 2023-10-31
chore(GHA): remove `on.*.paths` from required workflows

lemeurherve:
This PR removes the `on.*.paths` from workflows set as required status checks in branch protections.

These workflows already check if there are charts modified, and currently this prevents merging pull requests without bypassing branch protections if they don't contain changes to the `charts` folder.

#### Example of blocked PR where we can notice required status checks don't start:
- https://github.com/jenkins-infra/helm-charts/pull/912

image

#### Comparison with this PR:
image

### DIFF
--- a/.github/workflows/check-chart-versions-bump.yml
+++ b/.github/workflows/check-chart-versions-bump.yml
@@ -1,8 +1,6 @@
 name: Check chart version bump
 on:
   pull_request:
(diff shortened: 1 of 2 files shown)


dduportal (approved):
Approving to check the new behavior  as soon as posible

PR #1006, opened 2024-01-09
cleanup: remove legacy `mirrorbits` chart and rename `mirrorbits-lite` to `mirrorbits` 1/2

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/3828#issuecomment-1882509992

Note that `mirrorbits-parent` is not updated yet as it require a first release of the "new" `mirrorbits` chart (`2.0.0+`)


lemeurherve (approved):
🔥

dduportal:
Note: checks are failing but for expected reasons:

- Version bump checks fails with only `ERROR: the version of the 'mirrorbits-lite' chart hasn't been bumped: ''.` error (as we removed the chart)
- Updatecli reports:
  - 1 changed dependency for `mirrorbits-parent` (expected as we introduce a new version of at least one of its subcharts)
  - 1 failure when searching for the `mirrorbits` image tag  (expected as it is part of this change)

PR #1007, opened 2024-01-09
cleanup!(mirrorbit-parent) change dependency to support renaming of `mirrorbits-lite` to `mirrorbits`. BREAKING.

dduportal:
Follow up of #1006 since https://github.com/jenkins-infra/helm-charts/releases/tag/mirrorbits-2.0.0 has been released.

### DIFF
--- a/charts/mirrorbits-parent/Chart.lock
+++ b/charts/mirrorbits-parent/Chart.lock
@@ -1,12 +1,12 @@
 dependencies:
-- name: mirrorbits-lite
+- name: mirrorbits
   repository: https://jenkins-infra.github.io/helm-charts
-  version: 0.4.1
+  version: 2.0.0
 - name: httpd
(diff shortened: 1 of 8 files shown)


lemeurherve (approved):
LGTM

It might have been better to name the new version of `mirrorbits` 1.0.0 instead of 2.0.0 to avoid a small confusion... 🤷 🚢

dduportal:
> LGTM
> 
> It might have been better to name the new version of `mirrorbits` 1.0.0 instead of 2.0.0 to avoid a small confusion... 🤷 🚢

Not possible as the chart named `mirrorbits` already existed before and had the `1.0.0` version published: https://github.com/jenkins-infra/helm-charts/releases/tag/mirrorbits-1.0.0.

Since the "renaming" is a breaking change, then the version scheme increments the major digit leading to `2.0.0`

PR #1013, opened 2024-01-10
chore(mirrorbits-parent) fix unittest for helm-unittest `v0.4.1`

dduportal:
PR #1011 bumped the helm unittests plugin to [`v0.4.1`](https://github.com/helm-unittest/helm-unittest/releases/tag/v0.4.1).

This version bump made the unittest failing for `mirrorbits-parent` due to https://github.com/helm-unittest/helm-unittest/pull/267: we cannot use "test-scoped" values defined inline with the `set` keyword.

This PR uses a common values files which has the benefit to avoid scattering values over time AND can be overridden per sub test.

Note @lemeurherve , the PR 1011 did not run any check as it detected no chart modification. Do you think it is possible (and easy) to run all unittests if the file in charge of installing the plugin (such as the GHA workflow) is changed?

### DIFF
--- a/charts/mirrorbits-parent/Chart.yaml
+++ b/charts/mirrorbits-parent/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: mirrorbits-parent
 description: A mirrorbits parent chart for Kubernetes
 type: application
-version: 2.0.1
+version: 2.0.2
 maintainers:
 - email: jenkins-infra-team@googlegroups.com
   name: jenkins-infra-team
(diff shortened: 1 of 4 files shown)


lemeurherve (approved):

lemeurherve:
> Note @lemeurherve , the PR 1011 did not run any check as it detected no chart modification. Do you think it is possible (and easy) to run all unittests if the file in charge of installing the plugin (such as the GHA workflow) is changed?

@dduportal addressed in https://github.com/jenkins-infra/helm-charts/pull/1014

PR #1017, opened 2024-01-11
chore(updatecli): small typo in target name (visible in generated PRs)

lemeurherve:
See typo in https://github.com/jenkins-infra/helm-charts/pull/1016 for example:
image

PR #1020, opened 2024-01-13
fix(GHA/test): don't fail

lemeurherve:
This PR fixes the test.yaml GHA. Tested by cherry-picking #1017 then #1019 commits, both got their checks passing.

Supersedes and closes #1017 

Ref:
- https://github.com/jenkins-infra/helm-charts/pull/1017#issuecomment-1886623791

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,19 +18,20 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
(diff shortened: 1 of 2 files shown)


lemeurherve:
As #1017 has already been approved but is still failing on this GHA even after re-run its check, I've cherry-picked it here again for validating the fix.
(No idea why these changes work 🤷)

timja (approved):

PR #1014, opened 2024-01-10
chore(GHA/test): run tests on all charts if this workflow is modified

lemeurherve:
From #1013:
> [...] the PR #1011 did not run any check as it detected no chart modification. Do you think it is possible (and easy) to run all unittests if the file in charge of installing the plugin (such as the GHA workflow) is changed?

This PR fixes that.

PR #1022, opened 2024-01-15
fix(GHA/test): don't try to retrieve modified files from PR if not in a PR

lemeurherve:
Same player starts again... Follow-up of #1020

Tested by executing the script in local on #1014, #1017 and #1019


smerle33 (approved):
lets try

Issue #261, opened 2022-10-20
feat(jenkins-kubernetes-agent) disable token automount


dduportal:
Ref. https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: build-robot
automountServiceAccountToken: false
```

SarathChandra24:
@dduportal Can you close this issue if this [PR:](https://github.com/jenkins-infra/helm-charts/pull/1074) solved it.

dduportal:
> @dduportal Can you close this issue if this [PR:](https://github.com/jenkins-infra/helm-charts/pull/1074) solved it.

The PR is merged, but the chart is not deployed to the production: I'll wait for this before closing the issue :)

PR #1074, opened 2024-02-19
chart: jenkins-kubernetes-agent "disable token automount"

SarathChandra24:
Disable service account token automount for chart: jenkins-kubernetes-agent 

https://github.com/jenkins-infra/helm-charts/issues/261

### DIFF
--- a/charts/jenkins-kubernetes-agents/Chart.yaml
+++ b/charts/jenkins-kubernetes-agents/Chart.yaml
@@ -3,5 +3,5 @@ apiVersion: v2
 name: jenkins-kubernetes-agents
 description: A Helm chart for using Jenkins Kubernetes agents on a remote cluster
 type: application
-version: 0.5.2
+version: 1.0.0
 appVersion: "0.1.0"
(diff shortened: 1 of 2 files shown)


charts/jenkins-kubernetes-agents/Chart.yaml
```diff
@@ -3,5 +3,5 @@ apiVersion: v2
 name: jenkins-kubernetes-agents
 description: A Helm chart for using Jenkins Kubernetes agents on a remote cluster
 type: application
-version: 0.5.1
+version: 0.5.2
```
dduportal, line 6:
```suggestion
version: 1.0.0
```

This change is not backward compatible so it is a major bump (it means "if applied then the current behavior changes")

dduportal (changes requested):
Many thanks for this contribution!

I've added a suggestion about the version number: I'll let you review and add it, otherwise I'll merge it in 24h with my suggestion (if no answer back).

dduportal (approved):
Thanks! LGTM

PR #1085, opened 2024-02-26
fix(jenkins.io): don't hardcode `jenkinsio` resource name and references

lemeurherve:
While working on jenkins.io storage account replacement, I noticed that some refs are hardcoded, and prevent deploying this chart under another release name than `jenkinsio`. This PR fixes that.

Related:
- https://github.com/jenkins-infra/helpdesk/issues/3414#issuecomment-1961011896


lemeurherve:
Self-merging to continue on the subject.

PR #1119, opened 2024-04-05
feat(artifact-caching-proxy) tune proxy upstream buffering and enable gzip with upstream

dduportal:
This change is motivated by https://github.com/jenkins-infra/helpdesk/issues/3969#issuecomment-2027356553. The goal is to decrease or remove the warning messages:

```
an upstream response is buffered to a temporary file /var/cache/nginx/proxy_temp/<...> while reading upstream
````

It includes the 2 following elements:

- Increasing the proxy buffer size given the size of the files served by ACP. It maps to JFrog Artifactory recommended settings (https://jfrog.com/help/r/artifactory-how-to-enable-tls-within-the-jfrog-platform/artifactory-nginx-configuration).
    - Note: This will increase memory usage but the ACP metrics shows than there is enough memory to handle this change

- Enable gzip for requests with the upstream to decrease the size of in-memory required buffers of responses - Note: This will increase CPU usage but the ACP metrics shows that CPU is doing almost nothing.

### DIFF
--- a/charts/artifact-caching-proxy/Chart.yaml
+++ b/charts/artifact-caching-proxy/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 description: artifact-caching-proxy is a Nginx caching proxy in front of repo.jenkins-ci.org
 name: artifact-caching-proxy
 type: application
-version: 1.0.0
+version: 1.1.0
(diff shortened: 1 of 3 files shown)


dduportal:
Tested locally in a k3d cluster: I can run a BOM `prep.sh` build.
I saw a slight increase during the initial request warm-up (+5% CPU and +3% median response time from upstream to buffer) as expected, but no differences in memory usage or in behavior when artifacts are cached.
The `warn` message disappeared between my `1.0.0` test and `1.1.0`.

timja (approved):
Nice!

PR #1135, opened 2024-04-22
feat(httpd): allow to set a custom httpd conf

lemeurherve:
This PR allows to set a custom httpd conf.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2068909070


dduportal (approved):
Note: this change allows replacing the existing `httpd.conf`. It does NOT allows extending or partial override. Not a problem but worth making this explicit.

PR #1169, opened 2024-05-25
make first proxy target path configurable

sboardwell:
This PR makes the initial proxy location path configurable.

The chart version has been set, anticipating #1168 being merged first.


dduportal (approved):
👏

PR #1173, opened 2024-05-29
make dns resolver configurable

sboardwell:
This PR makes the dns resolver IP configurable since some people have their own DNS servers.

- Allow configuration of dns resolver
- Bump chart version for new feature


dduportal (approved):
👏

PR #1194, opened 2024-06-23
feat(jenkins-jobs): allow defining excluded branch pattern

lemeurherve:
This PR allows configuring excluded branch pattern for jobs.

Example of use-case: not running pipeline dedicated to PR previews on primary branch.

Ref:
- https://github.com/jenkins-infra/kubernetes-management/pull/5345


timja (approved):

dduportal (approved):
If it works (e.g. please either test it on a real Jenkins instance on your machine, or be available and check the result when this will be deployed to production)

PR #1268, opened 2024-08-14
feat!(mirrorbits) define mirrorbits.conf using (visible) values

dduportal:
It introduces a new `config` top-level value with a set of default values mapping to mirrorbit reference configuration. Implementation shifts from a Secret to a ConfigMap.

Note that this ConfigMap could hold a Redis password but RBAC allows reading it in clear if you can exec to the container so the secret does not protect more than ConfigMap.

BREAKING CHANGE: the value `conf` is deprecated and ignored. Check the new `config` value instead.
BREAKING CHANGE: the value `logs` is deprecated and ignored. Check the new `config.logs` value instead.

### DIFF
--- a/charts/mirrorbits/templates/NOTES.txt
+++ b/charts/mirrorbits/templates/NOTES.txt
@@ -2,10 +2,10 @@
 Check for deprecated values
 */}}
 {{- if .Values.conf }}
-[⚠️ WARNING] Found legacy value `conf` deprecated with the version 3.x of th chart. Please check the new value `config` instead.
+[⚠️ WARNING] Found legacy value `conf` deprecated with the version 3.x of the chart. Please check the new value `config` instead.
 {{- end }}
 {{- if .Values.logs }}
-[⚠️ WARNING] Found legacy value `logs` deprecated with the version 3.x of th chart. Please check the new value `config.logs` instead.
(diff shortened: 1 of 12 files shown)


charts/mirrorbits/templates/NOTES.txt
```diff
@@ -1,3 +1,13 @@
+{{/*
+Check for deprecated values
+*/}}
+{{- if .Values.conf }}
+[⚠️ WARNING] Found legacy value `conf` deprecated with the version 3.x of th chart. Please check the new value `config` instead.
```
timja, line 5:
```suggestion
[⚠️ WARNING] Found legacy value `conf` deprecated with the version 3.x of the chart. Please check the new value `config` instead.
```

charts/mirrorbits/templates/NOTES.txt
```diff
@@ -1,3 +1,13 @@
...
+{{- end }}
+{{- if .Values.logs }}
+[⚠️ WARNING] Found legacy value `logs` deprecated with the version 3.x of th chart. Please check the new value `config.logs` instead.
```
timja, line 8:
```suggestion
[⚠️ WARNING] Found legacy value `logs` deprecated with the version 3.x of the chart. Please check the new value `config.logs` instead.
```

timja (approved):

PR #1297, opened 2024-08-20
feat(nginx-website): allow to customise nginx configuration

lemeurherveCB:
This PR allows to customise nginx configuration so we can set for example a `try_files` directive, required for React applications for direct links. 

It also sets the port in deployment.yaml to `service.port` in order to use the same value in every template.

Note: I've kept current values to avoid breaking existing releases using this chart, these default values could be updated to more sensible ones like `server_name _;` in the future.

Ref:
- https://github.com/jenkins-infra/stats.jenkins.io/issues/118
- https://github.com/jenkins-infra/helpdesk/issues/4132#issuecomment-2298635602

### DIFF
--- a/charts/nginx-website/tests/defaults_test.yaml
+++ b/charts/nginx-website/tests/defaults_test.yaml
@@ -60,4 +60,4 @@ tests:
           pattern: 'autoindex on;'
       - notMatchRegex:
           path: data["default.conf"]
-          pattern: 'try_files $uri /index.html;'
+          pattern: 'try_files \$uri /index.html;'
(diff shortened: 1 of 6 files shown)


charts/nginx-website/tests/defaults_test.yaml
```diff
@@ -33,3 +33,31 @@ tests:
...
+          pattern: 'root   /usr/share/nginx/html;'
+      - matchRegex:
+          path: data["default.conf"]
+          pattern: 'index  index.html index.htm;'
+      - matchRegex:
+          path: data["default.conf"]
+          pattern: 'autoindex on;'
+      - notMatchRegex:
+          path: data["default.conf"]
+          pattern: 'try_files $uri /index.html;'
```
lemeurherveCB, line 63:
```suggestion
          pattern: 'try_files \$uri /index.html;'
```
Forgot to fix this (non-matching) regexp.

dduportal (approved):

PR #1303, opened 2024-08-22
fix(jenkins-jobs) GH App credential owner must be a groovy string

dduportal:
Ref. https://github.com/jenkins-infra/helpdesk/issues/4165#issuecomment-2304081117

The `owner()` function in a GH App credential JobDSL definition must be a string. Otherwise it is considered a Groovy reference and any special character is treated as a Groovy operator (such as the `-` of `jenkins-infra`...).


charts/jenkins-jobs/tests/credentials_test.yaml
hervelemeur, line 1: Late suggestion: use an org containing a dash in its name for tests.

dduportal: That would not have any sense as there are no Groovy interpolation happening in the unit tests. 
So it would not test anything.

PR #1313, opened 2024-08-28
chore(jenkins-jobs) code and metadata cleanup

dduportal:
- Use the properer maintainer in Chart.yaml
- Split template into different files per theme to avoid a big helpers file
- Cleanup indentation in template helpers
- Improve e2e manual testing (avoid crumb errors and add Jenkins default auth.)

----

Note: tried a `diff` on production: it only remove whitespaces and empty lines \o/

----

Note: about file naming conventions in Helm: https://helm.sh/docs/chart_template_guide/named_templates/#partials-and-_-files

> files whose name begins with an underscore (_) are assumed to not have a manifest inside. These files are not rendered to Kubernetes object definitions, but are available everywhere within other chart templates for use.

### DIFF
--- a/charts/jenkins-jobs/Chart.yaml
+++ b/charts/jenkins-jobs/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 description: Jenkins Jobs Definition from YAML to job-dsl
 icon: https://jenkins.io/images/logos/plumber/256.png
 maintainers:
-  - name: Damien Duportal <damien.duportal@gmail.com>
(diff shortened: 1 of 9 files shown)


smerle33 (approved):
as a review with you it seen clear and natural 👍

dduportal:
Note: might have caused infra.ci failure and was reverted.
More details here to follow up: https://github.com/jenkins-infra/helpdesk/issues/4165#issuecomment-2323979337

PR #1317, opened 2024-08-30
feat(jenkins-jobs) factorize Jenkins Infra default settings to values

dduportal:
This PR is a code cleanup to help the work on implementing Organization scanning (related to https://github.com/jenkins-infra/helpdesk/issues/4165).

It allows customizing the default GitHub Org, the default Jenkinsfile path and the default job trigger scan interval through a new top level value `defaults`. 
This behavior replaces having the defaults values written raw in the template.
It also allows overriding the values per job.

Notes:

- Unit test might not cover all cases (quick factorization)
- Checked against production: shows no changes in the helmfile diff
- We could need, ability to override the value per folder (like we do for GH scan credential ID). Not done (again: quick factorization)
- The credential unit tests had been simplified to limit the coupling with multibranch template code: changing the default values for GH owner is not failing the credential test anymore
  -  Under the hood: the test checks for specific lines, using (golang) regex, instead of matching the whole text. The trade off between "regexp" (e.g. PITA) and coupling is in favor of regex in this case.
  - Note: I failed to find a way to run unit tests only on a defined helper template. It's a limit of the helm unittest plugin which as an atomic level of template files, not template helper function.


timja (approved):

dduportal:
Note: might have caused infra.ci failure and was reverted.
More details here to follow up: https://github.com/jenkins-infra/helpdesk/issues/4165#issuecomment-2323979337

PR #1316, opened 2024-08-29
chore(jenkins jobs) improve tests and fix whitespaces (almost)

dduportal:
ref. https://github.com/jenkins-infra/kubernetes-management/pull/5633#pullrequestreview-2267950153

PR #1321, opened 2024-09-02
fix(jenkins-jobs) avoid error logs by ensuring no spaces before newlines

dduportal:
Related to https://github.com/jenkins-infra/helpdesk/issues/4165#issuecomment-2323979337

This PR is a fixup of both #1313 and #1316.

Unit tests are not enough to catch this issue, but end to end (manual) tests allowed me to reproduce the production error.
Might be worth adding e2e tests (at least here in a first step).

About the fix: when it comes to multi-line strings with YAML: https://yaml-multiline.info/ is the way.

Root cause of the errors is described in https://github.com/helm/helm/issues/5446#issuecomment-866994215:

> However removing any instance of a space before a newline fixed it.

by removing some of the whitespaces (pain with Helm :'(), I ended up into having spaces before empty lines which confused JobDSL.
It was visible by exec-ing on the Jenkins pod and showing the YAML file content which showed broken JobDSL, while the ConfigMap itself did not show it immediately.

This change had been applied on both the local end to end  (test from scratch) + on infra.ci (with a JCasC reload: no error message). Hoping it should be ok (but the fact it fixed the error on local e2e k3d is positive)

### DIFF
--- a/charts/jenkins-jobs/Chart.yaml
+++ b/charts/jenkins-jobs/Chart.yaml
@@ -4,5 +4,5 @@ icon: https://jenkins.io/images/logos/plumber/256.png
 maintainers:
   - name: Jenkins Infra Team <jenkins-infra-team@googlegroups.com>
 name: jenkins-jobs
-version: 2.2.0
+version: 2.2.1
 appVersion: "1.0" # unused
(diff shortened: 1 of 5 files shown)


lemeurherveCB (approved):
YAML \o/ 😅

PR #1331, opened 2024-09-12
feat(httpd/restart): provide a cronjob for rollout restart

smerle33:
following the last brownout from https://github.com/jenkins-infra/helpdesk/issues/2649
we saw few errors from httpd not handling properly the htaccess content. 
This new helm allow to enable an automatic rollout restart of the httpd deployment every hour to ensure httpd file content in pod memories.

we decided to provide a cronjob to restart the pods at pre-defined time (every hours for now), this will lower the impact of that bug. (see https://github.com/jenkins-infra/helpdesk/issues/2649#issuecomment-2352781184)

we need to keep in mind that we may need to provide a cleanup for both [cronjobs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#jobs-history-limits) and [jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically) through the `.spec.successfulJobsHistoryLimit` and `.spec.failedJobsHistoryLimit` fields for the first and the `.spec.ttlSecondsAfterFinished` field  for the second


charts/httpd/values.yaml
```diff
@@ -79,3 +79,10 @@ httpdConf:
 nodeSelector: {}
 tolerations: []
 affinity: {}
+httpdRestart:
+  enable: false # allow automatic kubernetes rollout restart
+  cron: '0 * * * *' # default to every hour
+  image:
+    repository: rancher/kubectl
+    tag: v1.28.13
```
dduportal, line 87: Since we use 1.29.x, WDYT about keeping this one by default, and then we'll had an updatecli manifest to track the version (based on the AKS version used on publick8s)?

smerle33: I planned to use the updatecli to update this right after this PR but you are right lets put directly the correct one.

dduportal: We're saying the same thing ;) We agree then

smerle33:
tested locally on a k3d install: 

```
NAMESPACE            NAME                                       SCHEDULE    SUSPEND   ACTIVE   LAST SCHEDULE   AGE
updates-jenkins-io   updates-jenkins-io-httpd-restart-cronjob   0 * * * *   False     0        <none>          2m54s
```

charts/httpd/templates/httpdRestart.yaml
```diff
@@ -0,0 +1,64 @@
...
+  kind: Role
+  name: {{ include "httpd.fullname" . }}-restart-deployment
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "httpd.fullname" . }}-restart-deployment
+---
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "httpd.fullname" . }}-restart-cronjob
```
dduportal, line 36:
```suggestion
  name: {{ include "httpd.fullname" . }}-restart-deployment
```

- Homogeneity with the other objects created withing the same functional scope (svcaccount, role, etc.)
- putting the type of the object in the name is usually duplicating information

charts/httpd/templates/httpdRestart.yaml
```diff
@@ -0,0 +1,64 @@
...
+  labels:
+{{ include "httpd.labels" . | indent 4 }}
+spec:
+  successfulJobsHistoryLimit: 1
+  failedJobsHistoryLimit: 2
```
dduportal, line 41:
```suggestion
```

You mentioned we could eventually fine tune later: in that case let's stick with the Kubernetes default and avoid surprises

> we need to keep in mind that we may need to provide a cleanup for both [cronjobs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#jobs-history-limits) and [jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/#clean-up-finished-jobs-automatically) through the .spec.successfulJobsHistoryLimit and .spec.failedJobsHistoryLimit fields for the first and the .spec.ttlSecondsAfterFinished field for the second

charts/httpd/templates/httpdRestart.yaml
```diff
@@ -0,0 +1,64 @@
...
+  concurrencyPolicy: Forbid
+  {{- with .Values.httpdRestart.cron }}
+  schedule: {{ toYaml . }}
+  {{- end }}
```
dduportal, line 45:
```suggestion
  schedule: {{ .Values.httpdRestart.cron }}
```

- The `.Values.httpdRestart.cron` is not yaml: we have to remove the `toYaml` template function
- Using the `with` structure is not useful here: I assume `schedule` keyword is mandatory and must be specified.

charts/httpd/templates/httpdRestart.yaml
```diff
@@ -0,0 +1,64 @@
...
+  jobTemplate:
+    spec:
+      backoffLimit: 2
+      activeDeadlineSeconds: 600
```
dduportal, line 49: (nit) are these default values in Kubernetes? If no, what are the default (and are they sane for this use case)

charts/httpd/templates/httpdRestart.yaml
```diff
@@ -0,0 +1,64 @@
...
+      template:
+        spec:
+          serviceAccountName: {{ include "httpd.fullname" . }}-restart-deployment
+          restartPolicy: Never
+          containers:
+            - name: kubectl-cron
```
dduportal, line 55:
```suggestion
            - name: kubectl-httpdRestart
```

(nit) WDYT about setting container name to the intent (instead of using the wrapper Kubernetes object name `cron`)?

dduportal (commented):
First wave of review. We are in the good direction but some chart element should be updated (a bit of naming, but also default values for some CronJob attributes)

charts/httpd/templates/httpdRestart.yaml
smerle33: The `backoffLimit` is set by default to 6.
`activeDeadlineSeconds` as no default value.

see https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup

I feel inconfortable having a job with no limit, thus setting those 2 ... maybe we can leave the backoffLimit to it's 6 default and have a bigger deadline but a rollout restart should be fast or there is a problem.

smerle33: as we agreed I change to use the default backoffLimit of 6 and no activeDeadlineSeconds, will see later if we need to implement it.

dduportal (approved):

PR #1332, opened 2024-09-12
feat(nginx-website): allow setting redirections

lemeurherveCB:
This PR allows defining redirections (with `return`, preferable in most cases to `rewrite`) so we can for example avoid breaking existing (old.)stats.jenkins.io URLs which pointed to files on GitHub but are not served by the new frontend, used by far more than we/I though initially.

Ref:
- https://github.com/jenkins-infra/helpdesk/issues/4132#issuecomment-2347009570

### Example of usage

values.yaml:
```
nginx:
  serverName: localhost
  slashLocation:
    root: /usr/share/nginx/html
    index: index.html index.htm
    autoindex: "on"
    ## Uncomment to enable try_files directive
    # tryFiles: $uri /index.html
  redirections: {}
    - location: /jenkins-stats
      returnedCode: 301
      destination: https://raw.githubusercontent.com/jenkins-infra/infra-statistics/gh-pages/
    - location: /plugin-installation-trend
      returnedCode: 301
      destination: https://raw.githubusercontent.com/jenkins-infra/infra-statistics/gh-pages/
    - location: /pluginversions
      returnedCode: 301
      destination: https://raw.githubusercontent.com/jenkins-infra/infra-statistics/gh-pages/
```

Resulting nginx config:
```
    server {
      listen       80;
      server_name  localhost;
      location / {
        root   /usr/share/nginx/html;
        index  index.html index.htm;
        autoindex on;
      }
      location /jenkins-stats {
        return 301 https://raw.githubusercontent.com/jenkins-infra/infra-statistics/gh-pages/$request_uri;
      }
      location /plugin-installation-trend {
        return 301 https://raw.githubusercontent.com/jenkins-infra/infra-statistics/gh-pages/$request_uri;
      }
      location /pluginversions {
        return 301 https://raw.githubusercontent.com/jenkins-infra/infra-statistics/gh-pages/$request_uri;
      }
    }
```

### DIFF
--- a/charts/nginx-website/values.yaml
+++ b/charts/nginx-website/values.yaml
@@ -61,3 +61,8 @@ nginx:
     autoindex: "on"
     ## Uncomment to enable try_files directive
     # tryFiles: $uri /index.html
+  redirections: {}
+    ## Example:
+    # - location: /plugin-installation-trend
+    #   returnedCode: 301
+    #   destination: https://raw.githubusercontent.com/jenkins-infra/infra-statistics/gh-pages/
--- a/charts/nginx-website/Chart.yaml
+++ b/charts/nginx-website/Chart.yaml
@@ -5,4 +5,4 @@ description: A Helm chart for serving website with nginx
 name: nginx-website
 maintainers:
 - name: lemeurherve
-version: 0.2.0
+version: 0.3.0
--- a/charts/nginx-website/templates/nginx-configmap.yaml
+++ b/charts/nginx-website/templates/nginx-configmap.yaml
@@ -16,4 +16,11 @@ data:
         try_files {{ .Values.nginx.slashLocation.tryFiles }};
         {{- end}}
       }
+      {{- if .Values.nginx.redirections }}
+      {{- range .Values.nginx.redirections }}
+      location {{ .location }} {
+        return {{ .returnedCode }} {{ .destination}}$request_uri;
+      }
+      {{- end}}
+      {{- end}}
     }
--- a/charts/nginx-website/tests/custom_values_test.yaml
+++ b/charts/nginx-website/tests/custom_values_test.yaml
@@ -11,6 +11,13 @@ set:
       index: custom.html
       autoindex: "off"
       tryFiles: $uri /index.html
+    redirections:
+      - location: /alocation
+        returnedCode: 301
+        destination: https://awebsite.example/apath/
+      - location: /anotherlocation
+        returnedCode: 302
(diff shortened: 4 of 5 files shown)


charts/nginx-website/values.yaml
```diff
@@ -61,3 +61,8 @@ nginx:
     autoindex: "on"
     ## Uncomment to enable try_files directive
     # tryFiles: $uri /index.html
+  redirections: {}
+    ## Example:
+    # - location: /plugin-installation-trend
+    #   returnedCode: 301
+    #   destination: https://raw.githubusercontent.com/jenkins-infra/infra-statistics/gh-pages/
```
timja, line 68: This isn't the same API and may still cause breaking changes.

1. There's no guarantee that consumers will follow redirects
2. raw.githubusercontent returns a content-type of text/plain which may break consumers

timja (commented):
This may work, but the safer option would be new stats app to server these files directly.

If it can't be done quickly then the site should be rolled back (although old should be kept to not breaking people who have migrated there)

lemeurherveCB:
> the safer option would be new stats app to server these files directly.

I agree with that and the other points you mentioned, but the issue with serving these files directly from our service is that we don't know (and can't AFAIK) how much outbound bandwidth it could represent.
Depending on the external consumers that might be too costly.

timja:
> > the safer option would be new stats app to server these files directly.
> 
> I agree with that and the other points you mentioned, but the issue with serving these files directly from our service is that we don't know (and can't AFAIK) how much outbound bandwidth it could represent. Depending on the external consumers that might be too costly.

If thats a concern is there any reason we moved off GitHub pages?
its a fully static site isn't it?

dduportal:
> This may work, but the safer option would be new stats app to server these files directly.
> 
> If it can't be done quickly then the site should be rolled back (although old should be kept to not breaking people who have migrated there)

+1 with @timja . This should justify a rollback and a re-architecturing of the model used if bandwidth is a problem.

dduportal:
> (although old should be kept to not breaking people who have migrated there)

@timja @lemeurherveCB Just a point about the GitHub Pages: it only accepts a single CNAME.
Looks like it is a problem here as we have to choose between `old.stats.jenkins.io` or `stats.jenkins.io` if we rollback.

I'll continue in https://github.com/jenkins-infra/helpdesk/issues/4132

dduportal:
Closed by https://github.com/jenkins-infra/stats.jenkins.io/pull/168 (not needed anymore)

PR #1347, opened 2024-09-18
chore(acp/report/acme): correct a type on {{- end }} missing a space

smerle33:
{{- end }} require space around the `end` word


dduportal (approved):
good catch!

PR #1910, opened 2026-04-15
feat(rating): use new healthcheck from 0.2.0 as `readinessProbe` path

lemeurherve:
Supersedes and close #1904 (updatecli force-pushed out my chart changes 🥲)

Ref:
- https://github.com/jenkins-infra/rating/pull/56


dduportal (approved):
Thanks!

PR #1959, opened 2026-06-04
cleanup(privatek8s-sponsored): remove `rss2twitter` chart

lemeurherve:
Ref:
- https://github.com/jenkins-infra/helpdesk/issues/5158


lemeurherve:
Note: as the "Check chart version bump" GHA is not taking in account chart removal, its failure is expected.

PR #1973, opened 2026-06-19
feat!(maven-cacher): allow to set release lines from chart values

lemeurherve:
This change will allow us to update those release lines without changing this chart code.

### DIFF
--- a/charts/maven-cacher/values.yaml
+++ b/charts/maven-cacher/values.yaml
@@ -28,7 +28,7 @@ resources: {}
 nodeSelector: {}
 tolerations: []
 affinity: {}
-cron: '0 14 * * 0' # Once a week, at 14:00, on Sunday
+cron: "0 14 * * 0"  # Once a week, at 14:00, on Sunday
 ## Provide a custom PVC to read and write caching archive from/to
 # cachePvc: customPersistentVolumeClaimName
 cacheMount: /cache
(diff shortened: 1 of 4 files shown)


charts/maven-cacher/values.yaml
```diff
@@ -28,7 +28,7 @@ resources: {}
 nodeSelector: {}
 tolerations: []
 affinity: {}
-cron: '0 14 * * 0' # Once a week, at 14:00, on Sunday
+cron: "0 14 * * 0"  # Once a week, at 14:00, on Sunday
```
dduportal: Why this change? Automatic formatting?

charts/maven-cacher/values.yaml
```diff
@@ -39,3 +39,5 @@ mavenMirror:
   enable: false
   # url: http://internal-mirror
   # mirrorOf: "external:*,!whatever"
+# Release lines included in jenkinsci/bom
+releaseLines: "weekly 2.504.x 2.516.x 2.528.x"  # space separated
```
dduportal:
```suggestion
releaseLines: "weekly"  # space separated
```

Since weekly will always exist:
- Pros: no confusion between default/not default value, no need to maintain this value
- Con: this would be, of course, a breaking change requiring the version bump to be `1.0.0` and the subsequent deployment on kubernetes to have the values updated (not an issue though).

dduportal (approved):

